PR #39, opened 2014-08-15
Make the raw body string available

richthegeek:
As far as I can tell, getting the raw body is currently impossible - using the "raw-body" module seems to return false regardless of what I do.

This puts it onto the req object should you want to use it directly instead of after parsing in cases where you need to do so (for example, HMAC checking)

### DIFF
--- a/lib/read.js
+++ b/lib/read.js
@@ -78,6 +78,9 @@ function read(req, res, next, parse, options) {
       next(err)
       return
     }
+    
+    // make this available outside of this module
+    req.rawBody = body
 
     // verify
     if (verify) {


charyorde:
Can you please provide some example?

I'm looking at `should error from verify` in https://github.com/expressjs/body-parser/blob/master/test/raw.js , but can't make a sense of it yet.

The question is how to use `buf` to make sure I get the raw request body (which is like this: `username=myname&password=mypass`) without having it transformed to json.

dougwilson:
@charyorde please open a new issue, with a lot more information for me to go on, rather than resurrecting a really old issue.

PR #75, opened 2015-01-07
Update README.md

efkan:


dougwilson:
I'm not sure how this addition really helps, because if we just one example, why not list out all the examples of what a rich object is?

PR #86, opened 2015-03-25
- a little hack to enable custom request decryption

vietanh85:
- Sometime we need to encrypt response and request from client to server to ensure security

### DIFF
--- a/lib/read.js
+++ b/lib/read.js
@@ -188,6 +188,7 @@ function decryptstream(input, debug, decrypt) {
   }
 }
 
+var Transform = stream.Transform || require('readable-stream').Transform
 /*
  * Decrypt an object stream
  *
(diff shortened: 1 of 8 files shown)


dougwilson:
Makes sense. Can the decryption be done as a stream instead of a sync decryption?

vietanh85:
Since, we are using raw-body to convert request stream to a buffer. In addition, I've learned that, there is no way to modify a readable stream in nodejs. So I think we cannot decrypt a stream.

dougwilson:
I'm talking about changing the interface here to take a TransformStream, rather than do it prior to calling this module.

dougwilson:
Ok, sorry we misunderstood. The _user_ needs to pass in a TransformStream as the decrypt argument. What we have here is still the same DoS vector.

dougwilson:
And if possible, please add tests and make sure decryption occurs after inflation.

Fatih86 (approved):

PR #87, opened 2015-03-29
Replace JSON.stringify(...) debug logging with %j

sehrope:
Replace explicitly calling `JSON.stringify(...)` when debug logging with `%j` so that `JSON.stringify(...)` is only called if logging is actually enabled.

This should save a function call in the normal case where debug logging for body-parser is not enabled.


dougwilson:
Looks like it still has the behavior I was looking for as well :+1:

PR #98, opened 2015-05-07
Allow extended options to be passed to qs library

Drarok:
This allows the options such as `arrayLimit` and `delimiter` to be passed to the `qs` library.

```js
app.use('/path', bodyParser.urlencoded({
  extended: {
    arrayLimit: 0,
    delimiter: ';'
  }
}));
```

The main benefit of this is that setting `arrayLimit` to 0 causes arrays to be represented as objects, allowing "sparse" arrays like the following:

```js
// foo[0]=1&foo[5]=hello
{ foo: { 0: 1, 5: "hello" } }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,14 +211,17 @@ any of the following keys:
 
 The `extended` option allows to choose between parsing the URL-encoded data
 with the `querystring` library (when `false`) or the `qs` library (when
-`true`). The "extended" syntax allows for rich objects and arrays to be
-encoded into the URL-encoded format, allowing for a JSON-like experience
-with URL-encoded. For more information, please
+`true`, or an object). The "extended" syntax allows for rich objects and
+arrays to be encoded into the URL-encoded format, allowing for a JSON-like
+experience with URL-encoded. For more information, please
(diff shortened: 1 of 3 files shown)


Drarok:
Tests pass locally, the error on Travis appears to be a timeout?

dougwilson:
> Tests pass locally, the error on Travis appears to be a timeout?

Seems to be some subtle change in the latest patch version of `on-finished` i.r.t Node.js 0.8. I'm' taking a look.

fatfisz:
We could allow all the other options, like `allowDots`, to be set too. Could I take over this?

dougwilson:
Because the `qs` module is officially been abandoned, we are going to switch to a different library, so we don't want to expose any options to `qs` through our API, otherwise we cannot easily switch to an alternative.

clayzermk1:
👍 It seems `qs` is alive and well today, any chance this could be re-examined?

solidgoldpig:
+1  Yes please to the reexamination

@dougwilson latest version of `qs` was published in April 2018

I'm confused as to whether `qs` is to be deprecated or not looking at the Release 2.0 PR.

PR #104, opened 2015-05-24
deps: bytes@~2.1.0

bachp:
This way body-parser and raw-body are using the same version.
This makes npm dedupe don't print a warning anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "license": "MIT",
   "repository": "expressjs/body-parser",
   "dependencies": {
-    "bytes": "1.0.0",
+    "bytes": "~2.1.0",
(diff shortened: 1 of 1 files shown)


dougwilson:
Thanks, though we are purposely holding off the update because it will require a minor version bump for this module, since it affects public API behavior (in that the limit option will be interpreted differently). Because of this and that the releases are still tied to express 3.x until it is dropped in July, I cannot promise when we will release a body-parser with the updated dep.

bachp:
Ok makes sense to keep this on hold then. Thanks for letting me know.

dougwilson:
Updated now :)! Will be released in the next minor, today.

PR #107, opened 2015-06-08
fix typo for JSON parsing in README

cvan:
`type` doesn't accept a regex. it uses `type-is` under the hood.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ All the parsers accept a `type` option which allows you to change the
 
 ```js
 // parse various different custom JSON types as JSON
-app.use(bodyParser.json({ type: 'application/*+json' }))
+app.use(bodyParser.json({ type: 'json' }))
 
 // parse some custom thing into a Buffer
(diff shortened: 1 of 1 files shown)


dougwilson:
Hi! That's not a RegExp; it's a string that `type-is` does indeed accept. Please see the last bullet point in the documentation: https://github.com/jshttp/type-is#each-type-can-be

cvan:
ah, okay. the `'application/*+json'` wildcard didn't work for me when I had a header of `Content-Type: application/json` - is that expected?

dougwilson:
Yes, because `application/json` doesn't have a suffix on that. It matches types like `application/x-custom+json`; if you wanted to match both then you can do `type: ['json', 'application/*+json']`, assuming you want to match the non-standard JSON types.

dougwilson:
To match `application/json`, you do not even specify the `type` option, as that is the default value (see https://github.com/expressjs/body-parser#type)

cvan:
gotcha. thanks for the clarification!

dougwilson:
No problem!

PR #116, opened 2015-08-24
Less confusing error message

GutsyGusty:
Was confused why this wasn't working when npm showed ISO-8859-1 was supported. This error message should prevent future confusion.


dougwilson:
Hi! It looks like this causes tests to fail.

Would you mind updating the tests to account for this change as well as documenting this new error message in the Errors sections in the readme :)?

GutsyGusty:
Sure. No problem.

GutsyGusty:
New PR is in for this.

PR #117, opened 2015-08-25
Better error message for json requests

codygustafson:
A new error will occur when a non-utf charset is attempted to be used for
requests with Content-Type: 'application/json'.

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -111,7 +111,7 @@ function json(options) {
     var charset = getCharset(req) || 'utf-8'
     if (charset.substr(0, 4) !== 'utf-') {
       debug('invalid charset')
-      next(createError(415, 'unsupported charset "' + charset.toUpperCase() + '"', {
+      next(createError(415, 'unsupported charset "' + charset.toUpperCase() + '". JSON must have a UTF charset', {
         charset: charset
       }))
       return
--- a/README.md
+++ b/README.md
@@ -311,6 +311,12 @@ This error will occur when the request had a `Content-Encoding` header that
 contained an unsupported encoding. The encoding is contained in the message
 as well as in the `encoding` property. The `status` property is set to `415`.
 
+### unsupported charset "BOGUS". JSON must have a UTF charset
+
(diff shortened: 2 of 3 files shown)


lib/types/json.js
```diff
@@ -109,9 +109,9 @@ function json(options) {
 
     // assert charset per RFC 7159 sec 8.1
     var charset = getCharset(req) || 'utf-8'
-    if (charset.substr(0, 4) !== 'utf-') {
+    if (['utf-8', 'utf-16', 'utf-32'].indexOf(charset) === -1) {
```
dougwilson: Hi! Please undo this line, because we allow the charset names UTF-16LE and UTF-16BE, and this will now deny those two charsets :(

GutsyGusty:
:grinning: No problem. Now updated.

dougwilson:
Awesome, thanks! I also just noticed that the new message no longer includes what the incorrect charset was in the message. We added that from a request just like this one :) Can the bad charset be included in the new message?

dougwilson:
On additional thought, it is likely to be confusing if you have "application/json" in the message, because people can change the type to anything they want, so hard-coding that type will mean that it may not match the type someone is actually using with the parser.

GutsyGusty:
Have changed this to just extend the original error message.

dougwilson:
Cool. So, I was just looking this over to see if it was merge-ready and there are just a couple minor issues still:

* This gives JSON a special case, because it was confusing that the charset was unsupported simply because it was the parser, not iconv-lite. Can you update the `urlencoded` parser to also have a similar message so they have parity?
* There is the text "but the iconv-lite module does not support it OR the parser does not support it", but with this change, there will be a specific error message for when the parser does not support the charset. Please update that text to remove the "OR" portion.

dougwilson:
Hi @codygustafson , I'm going to close this due to a lack of response to the outlined issues. If you want to make a new pull request when you get time to address them, that would be awesome. Otherwise, I'm going to close this for now instead of keeping it open indefinitely.

PR #121, opened 2015-09-04
Don't do dynamic require when it is not needed

### LINKED ISSUE #120: body-parser does not work with browserify
browserify is useful to package apps even at node level.

It can turn a 200MB dependency installation into a 2MB minified file.

The following code:
```
var parsersDir = path.join(__dirname, 'lib', 'types')

/**
 * Auto-load bundled parsers with getters.
 */

fs.readdirSync(parsersDir).forEach(function onfilename(filename) {
  if (!/\.js$/.test(filename)) return

  var loc = path.resolve(parsersDir, filename)
  var mod
  var name = path.basename(filename, '.js')

  function load() {
    if (mod) {
      return mod
    }

    return mod = require(loc)
  }

  Object.defineProperty(exports, name, {
    configurable: true,
    enumerable: true,
    get: load
  })
})
```
dynamically compute require input while there are only 4 modules to discover.

0xR:
Fixes #120

### DIFF
--- a/lib/types/urlencoded.js
+++ b/lib/types/urlencoded.js
@@ -19,18 +19,14 @@ var debug = require('debug')('body-parser:urlencoded')
 var deprecate = require('depd')('body-parser')
 var read = require('../read')
 var typeis = require('type-is')
+var qs = require('qs')
+var querystring = require('querystring')
 
 /**
  * Module exports.
  */
 
 module.exports = urlencoded
-
-/**
- * Cache of parser modules.
- */
-
-var parsers = Object.create(null)
 
 /**
  * Create a middleware to parse urlencoded bodies.
@@ -126,7 +122,7 @@ function extendedparser(options) {
   var parameterLimit = options.parameterLimit !== undefined
     ? options.parameterLimit
     : 1000
-  var parse = parser('qs')
+  var parse = qs.parse
 
   if (isNaN(parameterLimit) || parameterLimit < 1) {
     throw new TypeError('option parameterLimit must be a positive number')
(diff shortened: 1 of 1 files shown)


dougwilson:
The dynamic loading is on purpose for resource utilization for people who only use one of the other. This does not fix #120 completely, either. I am working on a fix that works with those builders and still keeps the lazy require feature (which we are not getting rid of).

dougwilson:
Actually, I had forgotten I already addressed the main module with https://github.com/expressjs/body-parser/commit/d6da231338e309c47adfae1f20146bbeb77895c4 ; the same kind of change is required in urlencoded, though it's weird the original reporter for that didn't flag it as still not working in browserify.

0xR:
When I look at d6da231 I don't think it's going to do anything for bundle size when using webpack or browserify. I presume the only goal is to reduce the memory footprint in plain node?

dougwilson:
> I presume the only goal is to reduce the memory footprint in plain node?

Correct. I don't think there is any way I can do anything about the bundle size for browserify or webpack, but as this is a Node.js module, it's primary support is for plain Node.js only, with other systems supposed as Class B when requested.

matianfu:
I like this PR. body-parser is the only module preventing me from webpacking the whole backend server. The memory footprint is no reason for unnecessary dynamic loading.

dougwilson:
I'm able to webpack this module just file. This PR was resolved a year ago. Perhaps you are using an old version of this module? If you are having issues, please open a new issue with enough details to help us help you. I'm locking this PR, as it was merged and you really should make a new issue.

PR #133, opened 2015-10-23
increase file size limit to 3GB

ocularcandy:
Update file size limit - this update was needed for the LD Grid Import update - Pat provided a spreadsheet with 800+ lines, this was crashing the importer due to file size limitations of 100kb

:boom:


dougwilson:
You need to update the limit in your own code. See the README for how to do so. Example: `app.use(bodyParser.json({ limit: '3000mb' }))`.

The default limits in the module protect servers against DoS attacks, as per security audits, so we don't be increasing the default limit, sorry.

PR #134, opened 2015-10-24
Add support for limit option as a function

sitegui:
This allows one to adjust the body size limit per-request, for example:

```js
app.use(bodyParser.json({limit: function (req, res) {
    if (req.url === '/import_something_huge') return '1mb'
    // uses default for everything else
}))
```

In our use case, we have an endpoint that imports a big (>100kb) chunk of data at once.
We've simply increased the limit for all requests for now, but I think it would be better if only the limit for that url was increased.

As far as I understand, the limit feature is here for security reasons (avoid DoS). And from the security point of view, there is no difference between increasing the limit of 1 endpoint or all them, since an attacker will try to exploit the weakest path. So this PR does not aim at increasing security, but it can't hurt it either.

What do you think?

PS: This PR does not update the docs, I just want to start the discussion around this idea, with some initial guess.

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -50,7 +50,7 @@ var firstcharRegExp = /^[\x20\x09\x0a\x0d]*(.)/
 function json(options) {
   var opts = options || {}
 
-  var limit = typeof opts.limit !== 'number'
+  var limit = typeof opts.limit !== 'number' && typeof opts.limit !== 'function'
     ? bytes.parse(opts.limit || '100kb')
     : opts.limit
   var inflate = opts.inflate !== false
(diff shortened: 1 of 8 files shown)


dougwilson:
You need to use the features of your framework (like Express). and mount the body parser at the URL you want, rather than add a redundant feature to this module. The most basic example if you are using a global `.use` statement (which is highly *not recommended*) is the following:
```js
app.use('import_something_huge', { limit: '1mb' })
app.use(bodyParser.json())
```

dougwilson:
The example above should hold you over, but eventually, I would suggest switching over to the pattern https://github.com/expressjs/body-parser#express-route-specific when you have the chance.

sitegui:
@dougwilson Thanks for the feedback ;)

In fact, I didn't know a global `.use` is so strongly not recommended.

dougwilson:
> As far as I understand, the limit feature is here for security reasons (avoid DoS). And from the security point of view, there is no difference between increasing the limit of 1 endpoint or all them, since an attacker will try to exploit the weakest path.

And as far as this comment goes, if you refactor your application into the suggested style, you should have your body parsing located after authentication and authorization, so then that route that needs to accept large bodies would have less of an attack surface (assuming, that is, that the general public Internet is not expected to be able to post to that route) :)

> In fact, I didn't know a global .use is so strongly not recommended.

Yea. Adding it as a top-level is so, so simple that everyone wants it, but bringing it down to be route-specific and after authentication/authorization, if possible, makes it even better, and of course, it's a bunch of work if it was done globally from the get-go, but i.m.o. a worth-while pattern.

Typically in most login-only applications, the only public body parser instance would be the login form, which could have a limit of `1kb`, or even less.

Issue #142, opened 2015-11-23
add a json5 parser


johanlelan:
I would like to be flexible on submitted JSON in request body. (allowing some mistakes)
To do this, i would like to use the [JSON5](https://github.com/aseemk/json5) parser instead of JSON one. I looked for an option to add external parsers but it seems that it does not exist.
What do you think about this new feature?

dougwilson:
Hi! This is a great suggestion and something we are tracking under #22

dougwilson:
Hi! This is a good idea, but we really don't want to keep pulling in more and more dependencies. We should really put the effort towards implementing #22 which would make implementing this in user-land or even as an add-on module trivial.

Issue #22, opened 2014-05-27
build-your-own-parser


dougwilson:
If you look at the middleware, after I had refactored it, all they are is a call to `typeis` and a call to `read`. We should add something like `bodyParser.generic()` to let people roll their owner simple body parsers.

lopugit:
How do you build may I ask?

jonchurch:
## Reviving this issue as a more central place to discuss all the Generic parser work.

### tldr; 
Im **in favor** of creating a **generic parser middleware** interface folks can extend to create custom middlewares for their advanced usecases. However, **Im against** exposing an `opts.parser` function on existing parsers which users can use as an escape hatch

## The Current PRs

We have several PRs open right now to land a Generic implementation. I don't want to leave feedback on all of these, or the additional closed ones, so Im writing this comment instead. cc @wesleytodd @UlisesGascon @ctcpip @Phillip9587 
https://github.com/expressjs/body-parser/pull/561
https://github.com/expressjs/body-parser/pull/575
https://github.com/expressjs/body-parser/pull/574

## Flexibility vs Simplicity

I'm open to offering a Generic interface for custom parsers, but I don’t believe we should expose an `opts.parser`moption in our existing middleware parsers. I prefer we export well known and well tested parsers which are hard to muck up and simpler to reason about for us and users. (We do that already with urlencoded extended, and I've read enough issues to know that confuses people.)

The goal of body-parser (IMO) is to provide simple, reliable parsers for common HTTP body formats like JSON and urlencoded data, covering common usecases therein. By sticking to common use cases and avoiding overcomplication, the core middleware is easier for users to adopt and maintain long term.

Extending a generic interface to create a more custom parser is a viable way to provide users flexibility while keeping the simplicity of the core middlewares.

Customization needs are rare compared to the more common issues reported, **which suggests that the current parsers cover most users' needs.**

We’ve seen a small but recurring set of issues where users want to extend or change the behavior of existing parsers. Examples include:

* Parsing BigInts https://github.com/expressjs/body-parser/issues/278
* Additional config options passed to `qs` https://github.com/expressjs/body-parser/pull/98 https://github.com/expressjs/body-parser/issues/453
* New formats such as XML, ndjson, or even YAML (!?) https://github.com/expressjs/body-parser/issues/370 https://github.com/expressjs/body-parser/issues/478 https://github.com/expressjs/body-parser/issues/369
* Streaming JSON parsers https://github.com/expressjs/body-parser/issues/132

[...]

PR #143, opened 2015-11-24
feat(JSON5): add JSON5 mode to parse body content

johanlelan:
pull request relating to #142


dougwilson:
Hi! This is a good idea, but we really don't want to keep pulling in more and more dependencies. We should really put the effort towards implementing #22 which would make implementing this in user-land or even as an add-on module trivial.

PR #163, opened 2016-03-16
options.type accepts array

heme:
Seems like your type property in options correctly passes an array on to type-is. Wasn't immediately obvious I could do that based on the README.

- I'm guessing other parsers should be able to do this (didn't change those)
- Not sure if the tests are needed as this seems to be a function of type-is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,12 +90,13 @@ accept anything `JSON.parse` accepts. Defaults to `true`.
 ##### type
 
 The `type` option is used to determine what media type the middleware will
-parse. This option can be a function or a string. If a string, `type` option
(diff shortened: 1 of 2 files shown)


dougwilson:
Awesome, thanks! Can you go ahead and also add it to all the other parsers as well? Currently I believe all their `type` sections are identical, so let's keep them identical.

dougwilson:
Hi @heme I'm going to close this due to a lack of response or anything. Let me know and I can re-open it, or you can always make a new pull request with the additional changes requested above :)!

shawninder:
I just made a PR that adds this to all other parsers as well: https://github.com/expressjs/body-parser/pull/268

Hopefully we can merge this in now?

PR #166, opened 2016-04-18
Add option for user to disable body-parser for some url

githubFree:
As I know, there is a simple way for us to use body-parser for one route,
But it seem to be very difficult for us to disable body-parser for some URL

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,10 @@ the following keys:
 When set to `true`, then deflated (compressed) bodies will be inflated; when
 `false`, deflated bodies are rejected. Defaults to `true`.
 
+##### disabled
+
+Disable body-parse for the url, This option should be a Regexp, Defaults to `false`.
+
 ##### limit
 
 Controls the maximum request body size. If this is a number, then the value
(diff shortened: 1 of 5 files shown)


dougwilson:
Hi @githubFree, thanks for your pull request! We appreciate the time you put into this, though wish you asked first :) The reason is that this type of request is always denied, because you should be performing this restriction using the standard middleware-wrapping pattern, rather than trying to get this added to every single middleware you care about. This is because you should be taking advantage of the middleware patterns.

A simple example of implementing your pull request as a wrapper pattern in your code:
```js
var bodyParser = require('body-parser')

function ignoreRoute(regexp, middleware) {
  return function (req, res, next) {
    if (regexp.test(req.url)) return next()
    return middleware(req, res, next)
  }
}
```

dougwilson:
Hi @githubFree, thanks for your pull request! We appreciate the time you put into this, though wish you asked first :) The reason is that this type of request is always denied, because you should be performing this restriction using the standard middleware-wrapping pattern, rather than trying to get this added to every single middleware you care about. This is because you should be taking advantage of the middleware patterns.

A simple example of implementing your pull request as a wrapper pattern in your code:
```js
var bodyParser = require('body-parser')

// ... you code

app.use(ignoreRoute(/^\/foo\//, bodyParser.json()))

// .. more code

function ignoreRoute(regexp, middleware) {
  return function (req, res, next) {
    if (regexp.test(req.url)) return next()
    return middleware(req, res, next)
  }
}
```

PR #168, opened 2016-04-20
json: update parse throw message to support node 6

### LINKED ISSUE #167: Quite a few tests breaking on node.js master / v6.0-rc
The v6.0 release is planned for next week and it would appear there are quite a few broken tests.

A few of them are related to changing error messages, but a few appear to not be as simple to fix. I am digging into it right now.

TheAlphaNerd:
Node v6 has some updated error messages. This commit allows the parse function to throw the correct error message for the appropriate version of node. It also updates the unit tests to check to using regular
expressions instead of strings in order to support the new error messages while remaining backwards compatible.

fixes #167

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -77,10 +77,13 @@ function json(options) {
 
     if (strict) {
       var first = firstchar(body)
-
+      var message = 'Unexpected token ' + first
       if (first !== '{' && first !== '[') {
+        if (process.version[1] >= 6) {
+          message += ' in JSON at position ' + (body.search(first) + 1)
+        }
         debug('strict violation')
-        throw new SyntaxError('Unexpected token ' + first)
+        throw new SyntaxError(message)
       }
     }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ node_js:
   - "3.3"
   - "4.4"
   - "5.8"
+  - "6.0"
 sudo: false
 before_install:
   # Setup Node.js version-specific dependencies
--- a/test/json.js
+++ b/test/json.js
@@ -23,7 +23,7 @@ describe('bodyParser.json()', function(){
     .post('/')
     .set('Content-Type', 'application/json')
     .send('{"user"')
-    .expect(400, 'Unexpected end of input', done)
+    .expect(400, /^Unexpected end of( JSON)? input/, done)
   })
 
   it('should handle Content-Length: 0', function(done){
@@ -62,8 +62,8 @@ describe('bodyParser.json()', function(){
     request(server)
     .post('/')
     .set('Content-Type', 'application/json')
-    .send('{:')
-    .expect(400, 'Unexpected token :', done);
+    .send(' {:')
+    .expect(400, /^Unexpected token :\s?(in JSON at position 2)?$/, done);
   })
 
   it('should 400 when invalid content-length', function(done){
@@ -117,8 +117,8 @@ describe('bodyParser.json()', function(){
       request(server)
       .post('/')
       .set('Content-Type', 'application/json')
-      .send('true')
-      .expect(400, 'Unexpected token t', done)
+      .send('    true')
+      .expect(400, /^Unexpected token t\s?(in JSON at position 5)?$/, done);
     })
 
     it('should allow leading whitespaces in JSON', function(done){
@@ -138,7 +138,7 @@ describe('bodyParser.json()', function(){
       .post('/')
       .set('Content-Type', 'application/json')
       .send('true')
-      .expect(400, 'Unexpected token t', done);
+      .expect(400, /^Unexpected token t\s?(in JSON at position 1)?/, done);
     })
   })
 


dougwilson:
Thanks, @TheAlphaNerd! It looks like the only issue is just tests failing vs. the module as a whole not working (*whew*). I really appreciate taking a look into this!

This is not a note for you, but just a note for the contributors here: we may want to take a look at making a change to better control the message being emitted rather than blindly passing back `JSON.parse` errors, not sure. Would like to hear some opinions.

TheAlphaNerd:
@dougwilson fwiw we are going to be doing a bunch of smoke testing on v6 this week. If this is not going to be able to land soon I'll likely add body-parser to the list of flaky modules (which I'd prefer not to do). Please let me know what would be best. 

Also, if you do not cut a new version we will also have to augment to test off of master, this is not a big deal for us, but just want to let you know that things will be running a bit differently.

dougwilson:
Hi @TheAlphaNerd, good to know. I'll try to get something done ASAP. As for the release thing, what does the smoke testing do when cloning? It sounds like it doesn't use `master`, but does it just use whatever the latest git tag is?

TheAlphaNerd:
@dougwilson mostly. It derives the semver of the tag based on whatever the latest published version is to npm. we can over ride that with master though, but prefer not to.

dougwilson:
@TheAlphaNerd good to know :) I usually don't like to make a random release, especially when no actual code was changed, but there are some outdated dependencies, so I can get a new version published for ya :)

jonathanong:
would love if all errors had a `.code` on it. not sure we can do that as a module or something without annoying everyone

TheAlphaNerd:
@dougwilson any movement on this?

dougwilson:
Yea, sorry. Once I got v6 working, I got a little side-tracked with some issues in the modules that underlay this one that have actual issue (where this is mostly a test issue, so it does take lower priority).

I can just merge this PR if you can make either one of the following changes:

1. Fix the JSON error messages so that they continue to match those from older node.js versions OR
2. Fix the strict violation message we throw in the JSON parse so that it matches the new message formation in v6 and continues to match the old message format in the old versions.

I did make a little head-way towards the first, but got caught up trying to fix some non-test incompatibilities in other modules for v6.

TheAlphaNerd:
@dougwilson the regular expressions used in these tests cover all old cases and the new cases.

When you are saying "fix the JSON error messages" are you inferring making changes to the body-parser source?

dougwilson:
Correct, there is a spot we are mimicking the native exception as a feature:

    throw new SyntaxError('Unexpected token ' + first)

This will be a broken feature for 6.0 and higher and imo consider correcting this part of making this module "work" on 6.0, in respect to promised features.

dougwilson:
I of course regret passing through these native errors and even trying to mimic them, as we are in a really crappy situation now. The first solution is the best one, which is why I was working on it, as these messages we have been giving out directly to the HTTP client. The reason we did a strict string match in our tests is because we have been promising that those are the exact error messages we provide. Changing the messages is effectively a semver major for this module, so the most ideal change would be to change our source to preserve the old messages while we evaluate how to better deal with this in the next major.

dougwilson:
TL;DR we have to make some change to our source for this. The two different ideas I have are listed in that numbered listed above.

TheAlphaNerd:
@dougwilson I've updated the PR and I believe it is now doing what you suggested in 2.

This is not the best solution IMHO, but as you said, changing errors is going to be a SEMVER MAJOR.

This does a very naive version check `process.version[1] >= 6` and then uses that result to generate the error message. The comparison only happens in the case where there is an error, so I don't think we should see any perf degredation here. 

I couldn't see any other instances of using the process object, or for doing version checking, so you might want to approach this a different way. 

Let me know if I'm off base, or if you require any other changes. My local tests pass on all major release streams

TheAlphaNerd:
I'm noticing that this has decreased coverage due to creating a code path that won't be followed if the version is different. This should be easy enough to write a test for if we cache the value of version in the module and overwrite it to test the two outcomes... usually I would use rewire for this kind of task, how have you instrumented this kind of stuff before? Does it even make sense to test robustly?

lib/types/json.js
```diff
@@ -77,10 +77,13 @@ function json(options) {
 
     if (strict) {
       var first = firstchar(body)
-
+      var message = 'Unexpected token ' + first
       if (first !== '{' && first !== '[') {
+        if (process.version[1] >= 6) {
+          message += ' in JSON at position 1'
```
dougwilson: Why is the position static? The character can be at any position, and this message would be confusing when it tells you that a double quote character is unexpected at position 1, but the first quote is 20 characters from the start of the string.

TheAlphaNerd: I was under the impression this throw was only for the first character based on the boolean logic just above

dougwilson:
The coverage is the sum of all tests on Travis CI. If you want it to keep the coverage the same, you can add 6.0 to the .travis.yml file, and that branch will end up covered.

lib/types/json.js
dougwilson: No, it throws based on the first non-ignored character. You can look at where that variable is actually coming from.

TheAlphaNerd: dawning on me that I misunderstood what firschar did...

could we replace the static number with `body.search(first)`?

TheAlphaNerd:
@dougwilson I've added v6.0 to the travis file. I did notice the versions in there are a bit out of date. Let me know if you would like a PR against master to update those.

I've updated the code based on the comment you left in line. It should dynamically report the placement of the character in the body now

TheAlphaNerd:
It is also dawning on me right now that there is a logical problem in my regular expressions... I'll comment in line

TheAlphaNerd:
I've updated things one more time. The regular expressions now have an anchor tag so we are not going to get false positives. I've also moved the characters in both examples to not be in the default character position of 1... that way we can test that we are giving the right character position to the developer. The two modified tests are using different interfaces for the error... one is the natural interface exposed by v8, the other is the message that is being created by `body-parser`. Doing this actually found a fun little bug in the intial code as I was zero indexing rather than 1 indexing the position of the character in the string

dougwilson:
Hi @TheAlphaNerd, sorry this took a while. With the discussion here, then done again internally and with some other logistics going on in Express at the moment, it just took some time to come back :) Updates to the tests should be in master now, with a release coming shortly (since you said the testing is based off releases and not just `master`). Sorry for the delay :S

PR #172, opened 2016-05-15
Accept charset like utf8, UTF8 and UTF-8

asvdw:
Some clients use http header "Accept-Charset: utf8" instead of "utf-8". This simple hack consider them as valid.


dougwilson:
Hi @asvdw, this module never reads the `Accept-Charset` header.

Issue #187, opened 2016-07-26
raw() parser: if body is empty, req.body is set to empty object


Olivier-OH:
I think the correct behaviour should be to have an empty string body.

dougwilson:
Hi @Olivier-OH, defaulting to a string in the raw parser does not make sense, since a successful raw parse results in a `Buffer` object, not a string. Besides, we actually went ahead and fixed these semantics in the 2.0 branch. Please go ahead and check out the proposed 2.0 release in #66 :)

molomby:
@dougwilson: You're right that defaulting to a string doesn't make sense but the issue @Olivier-OH reported is still valid.

The [docs](https://www.npmjs.com/package/body-parser#bodyparserrawoptions) indicate the raw body parser should always produce a body object that is a Buffer. In the current version (v1.15.2) this is not the case.

Specifically, when the raw parser is used, requests with an empty/zero-length body (eg. `GET` requests), result in the body object being an empty `Object`.

Surely, in this case, the body should be an empty `Buffer`.

PR #66, opened 2014-10-30
Release 2.0

dougwilson:
This is a tracking issue for release 2.0.

I am trying to give better visibility for upcoming changes and so am trying out making a PR for a release here, pulling from the official next release branch into `master`. This allows for the current pending changes to be easily visible.

**Please keep feature requests in their own issues**

I'm also leaving this PR unlocked so people can make comments/etc. and we'll see how it goes :) If you want to make a comment on a particular change, please make the comment in the "Files changed" tab so comments are not lost during a rebase.

List of changes for release:
- [x] Add pending change (@UlisesGascon) https://github.com/expressjs/body-parser/pull/66#pullrequestreview-2201522426
- [ ] ~Add generic body parser (@jonchurch @ctcpip TBD status)~
- [ ] ~Add `limit: null` as alias for `limit: Infinity` #493~
- [ ] ~Add multipart field body parser~
- [x] Drop support for Node.js 0.8
- [x] Drop support for Node.js below 18
- [x] `req.body` no longer always set to `{}`, which works better for non-object-like bodies like text
- [x] `on-finished` used to detect if the request has been read; no longer need to set `req._body` to a truthy value
- [x] `urlencoded` no longer defaults `extended` to `true`
- [x] Update `raw-body` to v3 https://github.com/stream-utils/raw-body/pull/87
- [x] #406 

List of deprecated things removed:
- [x] Main `bodyParser()` function which did `urlencoded` + `json`

**Testing this release**

If you want to try out this release, you can install it with the following command:

``` bash
$ npm cache clean
$ npm install expressjs/body-parser#2.x
$ npm cache clean
```

PR #188, opened 2016-07-26
Fix for raw() parser if body is empty

Olivier-OH:
Set it to empty string rather than an empty Object.

See issue https://github.com/expressjs/body-parser/issues/187


Olivier-OH:
I don't know whether I should change the unit test code. Is this important that the body is an empty object instead of an empty string?

Olivier-OH:
I changed the unit test code. I'll let the repo owners decide whether this is correct.

dougwilson:
Hi @Olivier-OH, defaulting to a string in the raw parser does not make sense, since a successful raw parse results in a `Buffer` object, not a string. Besides, we actually went ahead and fixed these semantics in the 2.0 branch. Please go ahead and check out the proposed 2.0 release in #66 :)

Olivier-OH:
Great! Thanks for your feedback.

Issue #42, opened 2014-08-26
Options must allow to provide options to queryparser


lazdmx:
Lib must allow to provide options to `queryparser` since there are cases when we need to provide more than 20-element array in query-string which by default will be parsed into JSON-Object instead of array.

![image](https://cloud.githubusercontent.com/assets/1035026/4040936/fd8e68aa-2cee-11e4-82e3-7367b4e732d5.png)

PR #203, opened 2016-10-20
Added support for custom urlencoded parsers

AlexMeah:
We are looking for a way to pass custom options to the urlencoded parser, this has previously been denied due to `qs` being deprecated https://github.com/expressjs/body-parser/pull/98.

Would this solution to allow custom parsers be something you'd be open to?

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,129 @@ This error will occur when the request had a `Content-Encoding` header that
 contained an unsupported encoding. The encoding is contained in the message
 as well as in the `encoding` property. The `status` property is set to `415`.
 
+### bodyParser.generic(options)
+
+Returns middleware that only parses `generic` bodies. This parser accepts
+any encoding of the body and supports automatic inflation of `gzip`
+and `deflate` encodings.
+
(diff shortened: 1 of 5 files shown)


dougwilson:
Hi @AlexMeah , I think the idea is in the right direction :) Instead of baking it in to be specific to the `urlencoded` parser, it would probably benefit a wider audience (including your use-case, to implement something more general, in line with the #22 idea. Some relevant conversation in #42

AlexMeah:
@dougwilson I'll hopefully get some time this week to make the changes

dougwilson:
Cool :) ! I started this work a while ago, on the `2.x-refactor` branch (https://github.com/expressjs/body-parser/tree/2.x-refactor) so you'll probably want to use that as a starting point to build it. The idea is that the module will simply be a pluggable generic parsing module and then the 4 parsers that already exist will simply all directly use the generic parsing (kind of like eat your own dog food: if all the existing parsers are just using the generic one, then it's a good indicator that the generic one is really generic and is a good base for everyone to use).

Issue #122, opened 2015-09-07
Difficult to determine that an error is the result of bad JSON


carlbennettnz:
Having `SyntaxErrors` consistently returned (#119) and a status code provided means we can fairly reliably detect errors in express middleware that are the result of bad JSON.

``` javascript
app.use(function(err, req, res, next) {
  if (err instanceof SyntaxError && err.status === 400) {
    console.error('Bad JSON');
  }
});
```

This test could theoretically give false positives though, if some other middleware were to throw something similar. Imagine we're using `body-parser` first and then something else to validate the body has the structure we're expecting, for example.

Is there any way to detect JSON parsing error with 100% reliability with the library as it is now? Is there something you could (perhaps in 2.0) to make this a little easier?

dougwilson:
AFAIK:
```js
app.use(function(err, req, res, next) {
  if (err instanceof SyntaxError && err.status === 400 && 'body' in err) {
    console.error('Bad JSON');
  }
});
```

carlbennettnz:
True. Still seems messy though.

dougwilson:
Suggestions :) ?

carlbennettnz:
Off the top of my head:

- Custom `Error` type
- Add a `_isJsonParsingError` property or something like that to the `SyntaxErrors`
- Some sort of error callback passed into `.json()`

dougwilson:
> Custom Error type

Probably the best, though without access, you cannot do `instanceof` against it.

> Add a _isJsonParsingError property or something like that to the SyntaxErrors 

We already do this, we just call the property `body` right now, and you didn't like that above, so weird to propose it anyway :)

> Some sort of error callback passed into .json() 

That wouldn't make any sense, because it's not how Express middleware works. You can easily wrap the body parsing middleware if you really want to know the error came from this middleware, and is the recommend way to determine error source:
```js
app.use(errorFork(bodyParser.json(), function (err, req, res, next) {
  // do stuff with only body parser errors
}))

// this is an example; you can use any pattern you like.
function errorFork(middleware, errorHandler) {
  middleware(req, res, function (err) {
    if (err) return errorHandler(err, req, res, next)
    return next()
  })
}
```

Any other suggestions :) ? I don't see anything to act on currently from that list.

carlbennettnz:
> Probably the best, though without access, you cannot do instanceof against it.

True, hadn't thought of that. You could expose it though I guess.

[...]

PR #206, opened 2016-11-17
Add option to validate the body as a json

rodrigowirth:
This is a proposal for the issue #122


dougwilson:
Hi @rodrigowirth it is not clear what this PR is even doing, even reading the docs and source. Can you give a much, much more detailed explanation of exactly what this PR does? It looks like, to me, all it does is prepend a word to the error message with the use of a switch?

dougwilson:
Oh, nevermind, I see the `res` change. Unfortunately, no, this PR is not going to be accepted. We ujse the standard error handing from Express and middleware; the module is never going to have functionality to write to the `res` object.

PR #215, opened 2016-12-21
Update qs dependency to 6.3.0

ysangkok:
The API has not significantly.

Changelog here: https://github.com/ljharb/qs/blob/master/CHANGELOG.md#630


ysangkok:
@dougwilson OK, please tell me if it is OK now.

dougwilson:
Please keep the PR to only one change. I also still don't see any description of that users will experience from the dep update in the history file.

dougwilson:
Sorry I don't have docs on what we expect; historically we have not accepted these kind of changes if it was simply to bump a dep. I will try to come up with some docs over this holiday season. No ETA on that or a merge, though.

ysangkok:
@dougwilson So, I reverted the other deps upgrades and added details on the changes I deem most important (new features added). The other changes are bug fixes or optimization, or refactorings that shouldn't be visible to users. Please tell me if this is in a mergeable state. I do not care much when this is merged, I just want the pull request to be perfect. Thanks in advance.

dougwilson:
Great, thanks! Can you remove the links and keep the descriptions concise? Use the other qs upgrades in the history file as an example. I am on and off planes, so cannot give more specific guidance for some time.

HISTORY.md
```diff
@@ -1,3 +1,11 @@
+1.15.3 / 2016-12-xx
```
dougwilson: This should just be "unreleased" you can use past commits to see how that is done until we have docs on the process.

HISTORY.md
```diff
@@ -1,3 +1,11 @@
...
+===================
+
+  * deps: qs@6.3.0
+    - Date stringification format configurable (https://github.com/ljharb/qs/issues/159)
```
dougwilson: We don't use the stringification API, so this is an irrelevant change to our users. Please only list the changes that affect our users.

HISTORY.md
```diff
@@ -1,3 +1,11 @@
...
+    - RFC 1738 support (`+` instead of `%20`). Must be explicitly enabled.
```
dougwilson: Is this a change in parsing or stringification?

HISTORY.md
```diff
@@ -1,3 +1,11 @@
...
+    - For remaining changes, see [qs/CHANGELOG.md](https://github.com/ljharb/qs/blob/master/CHANGELOG.md#630)
```
dougwilson: So the purpose of adding the information here is specifically so no one ever has to go reference the dependency change log :) I would remove this.

ysangkok:
@dougwilson The first two changes I listed here are not relevant for the library consumer. In fact, as far as I can see, the only change that change the result of functions is this one: `[Fix] ensure key[]=x&key[]&key[]=y results in 3, not 2, values`. Other changes are performance related. The other `[Fix]` doesn't seem to be actually relevant to the consumer of `qs`. Do you think it would be appropriate to just copy that one fix description and have nothing else?

PR #221, opened 2017-02-03
Attribute nesting

danielrohers:
Added option to perform attribute nesting on `req.body` or `req.query`.
If you have submitted `user.first_name` and `user.last_name`, an object is created.

Current:
```js
  req.body['user.first_name'] = 'daniel';
  req.body['user.last_name'] = 'moura';
```

New:

```js
  req.body.user.first_name = 'daniel';
  req.body.user.last_name = 'moura';
```


dougwilson:
Thanks for the PR, but I think it misses the goal of this module, namely parsing the request body. You can simply publish your type file as it's own middleware to npm.

PR #250, opened 2017-06-22
Avoid `throw` for an acceptable situation

JasonKleban:
When debugging and breaking on all exceptions, this one kept tripping because there is no charset parameter for my requests (for some reason) which seems reasonable.  Prechecks for that case and avoids the throw, still returning undefined as it would have.

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -154,7 +154,7 @@ function firstchar (str) {
 
 function getCharset (req) {
   try {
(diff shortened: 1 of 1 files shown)


dougwilson:
Sorry I didn't see your commit push earlier; looks like just a lint error failing the build, I can fix that & merge.

dougwilson:
P.S. I ported you change to the other parsers that were doing the same thing.

PR #259, opened 2017-09-02
Explicate optional parameter for parser methods

govindrai:
Here is a snippet from the [body-parser middleware documentation](https://expressjs.com/en/resources/middleware/body-parser.html):

> #  **bodyParser.json(options)**
> 
> Returns middleware that only parses json and only looks at requests where the Content-Type header matches the type option. This parser accepts any Unicode encoding of the body and supports automatic inflation of gzip and deflate encodings.
> 
> A new body object containing the parsed data is populated on the request object after the middleware (i.e. req.body).
> 
> **Options**
> 
> The json function takes an option `options` object that may contain any of the following keys:

The `options` objects that the JSON parser takes is optional, so shouldn't the snippet actually look like this?:

> #  bodyParser.json(**[**options**]**)
> 
> Returns middleware that only parses json and only looks at requests where the Content-Type header matches the type option. This parser accepts any Unicode encoding of the body and supports automatic inflation of gzip and deflate encodings.
> 
> A new body object containing the parsed data is populated on the request object after the middleware (i.e. req.body).
> 
> **Options**
> 
> The json function takes an **optional** `options` object that may contain any of the following keys:

Here are the two changes proposed
1. Add brackets around `options` parameter in the the method signature to signify optional `options` object
2. Change option `options` object to optional `options` object (last sentence)

I have applied the proposed changes to all four parsers (JSON, Raw, Text, URL-encoded).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ or an error occurred.
 The various errors returned by this module are described in the
 [errors section](#errors).
 
-### bodyParser.json(options)
+### bodyParser.json([options])
 
 Returns middleware that only parses `json` and only looks at requests where
 the `Content-Type` header matches the `type` option. This parser accepts any
@@ -71,7 +71,7 @@ object after the middleware (i.e. `req.body`).
 
 #### Options
 
-The `json` function takes an option `options` object that may contain any of
+The `json` function takes an optional `options` object that may contain any of
 the following keys:
 
 ##### inflate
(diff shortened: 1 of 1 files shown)


dougwilson:
> If so, similar changes can be applied to all four parsers (JSON, Raw, Text, URL-encoded).

Yea, go ahead and apply to all 4.

govindrai:
I have applied them to all four. Thanks.

dougwilson:
Oh, sorry.l, I thought you were asking a question about the change and that you only applied it to tge one and then were saying you can apply to to the rest. Can you clarify your initial post, a little confused now.

dougwilson:
When I get to a computer I'll checkit out. To clarify: the current changes in the PR is the final iteration, so there is not any open question from the first post, right?

govindrai:
you are correct. It was a question in my initial post on expressjs.com repo but when I saw your comment to just create a pr, I ended up creating the pr with the proposed changes and didn't revise the initial post. I have revised it now so it should make sense :)

PR #265, opened 2017-09-15
Allow user to pass extra options to `qs` parser

mdzzohrabi:
Allow user to pass extra options to `qs` parser, like `parseArrays` option


dougwilson:
See #58 and other discussions around this.

PR #275, opened 2017-10-27
Fixing a typo in the Readme

gayanhewa:
Fixing a typo in the Readme.md


dougwilson:
Hi @gayanhewa thank you for the PR! It's actually meant to be `res.end`, so it's not a typo in this case. Sorry for any confusion. The example is for Express/connect (as titled) and Connect does not have `res.send`, but they both have `res.end`.

gayanhewa:
@dougwilson thanks for clarifying.

PR #280, opened 2017-11-20
fix: " should be a valid first char in json documents

### LINKED ISSUE #279: sending stringified string make the parser fail
hi, if the payload of a POST request is created in this fashion it will make the parser crash:

```typescript
{
  body: JSON.stringify('BODY')
}
```

a stringified string won't be recognized as valid json due to this line: https://github.com/expressjs/body-parser/blob/master/lib/types/json.js#L81
`"` should be a valid first char. i will send a PR to fix that after verifying locally that is working.

phra:
fixes #279

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -78,7 +78,7 @@ function json (options) {
     if (strict) {
       var first = firstchar(body)
 
-      if (first !== '{' && first !== '[') {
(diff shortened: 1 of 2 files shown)


dougwilson:
This is already supported: set the option strict: false.

PR #281, opened 2017-11-21
Added support for external parsers to bodyParser.json()

### LINKED ISSUE #278: Any plans to support larger than 53-bit integers?
Integers larger than 53 bits are valid JSON, but when parsed by JSON.parse they get truncated down to 53 bits and lose a few least significant digits. We have 64 bit integers coming from our backend and use https://www.npmjs.com/package/json-bigint to parse, any chance of adding an option for replace the built-in JSON.parse with a custom function, or just a boolean option for using json-bigint?

shawndellysse:
Added the option to use a function other than `JSON.parse` to do the actual parsing. This fixes #278; contrary to one of the suggestions in that issue I left the `reviver` option as-is for backward compatibility.  Also see #42, where adding a similar option for query string parsing was the preferred choice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,16 @@ The `verify` option, if supplied, is called as `verify(req, res, buf, encoding)`
 where `buf` is a `Buffer` of the raw request body and `encoding` is the
 encoding of the request. The parsing can be aborted by throwing an error.
 
+##### parser
+
+The `parser` option, if supplied, is used to in place of the default parser to
+convert the request body into a Javascript object. If this option is supplied,
+both the `extended` and `parameterLimit` options are ignored.
+
+```javascript
+parser(body) -> req.body
+```
+
 ## Errors
 
 The middlewares provided by this module create errors depending on the error
--- a/README.md
+++ b/README.md
@@ -89,12 +89,18 @@ to `'100kb'`.
 ##### parser
 
 The `parser` option is the function called against the request body to convert
-it to a Javascript object. Defaults to `JSON.parse`.
+it to a Javascript object. If a `reviver` is supplied, it is supplied as the
+second argument to this function.
+
+```javascript
+parser(body, reviver) -> req.body
+```
+
+Defaults to `JSON.parse`.
 
 ##### reviver
 
-The `reviver` option is passed directly to the `parser` as the second
-argument. You can find more information on this argument
+You can find more information on this argument
 [in the MDN documentation about JSON.parse](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#Example.3A_Using_the_reviver_parameter).
 
 ##### strict
--- a/README.md
+++ b/README.md
@@ -170,6 +170,15 @@ The `verify` option, if supplied, is called as `verify(req, res, buf, encoding)`
 where `buf` is a `Buffer` of the raw request body and `encoding` is the
 encoding of the request. The parsing can be aborted by throwing an error.
 
+##### parser
+
+The `parser` option, if supplied, is used to transform the body of a request
+before being set to `req.body`.
+
+```javascript
+parser(body) -> req.body
+```
+
 ### bodyParser.text([options])
 
 Returns middleware that parses all bodies as a string and only looks at
--- a/README.md
+++ b/README.md
@@ -219,6 +219,15 @@ The `verify` option, if supplied, is called as `verify(req, res, buf, encoding)`
 where `buf` is a `Buffer` of the raw request body and `encoding` is the
 encoding of the request. The parsing can be aborted by throwing an error.
 
+##### parser
+
+The `parser` option, if supplied, is used to transform the body of a request
+before being set to `req.body`.
+
+```javascript
+parser(body) -> req.body
+```
+
 ### bodyParser.urlencoded([options])
 
 Returns middleware that only parses `urlencoded` bodies and only looks at
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ The `parser` option is the function called against the request body to convert
 it to a Javascript object. If a `reviver` is supplied, it is supplied as the
 second argument to this function.
 
-```javascript
+```
 parser(body, reviver) -> req.body
 ```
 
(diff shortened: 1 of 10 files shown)


lib/types/json.js
```diff
@@ -80,13 +81,13 @@ function json (options) {
...
         debug('strict violation')
-        throw createStrictSyntaxError(body, first)
+        throw createStrictSyntaxError(parser, body, first)
       }
     }
 
     try {
       debug('parse json')
-      return JSON.parse(body, reviver)
+      return parser(body, reviver)
```
dougwilson: I'm not sure this is right for custom parsers. Or at least it's not clear in the docs for the parser option that this is going to happen and how people should build their parser dunxtion to compensate.

shawndellysse: The other option would be to remove the `reviver` param and rely on people using that parameter should change to this. I didn't do that as that would break BC, and I figure a project like this wouldn't want to break BC very often.

Maybe the right way is to make the docs more clear about this?

dougwilson: Sorry, I commented on the wrog line.

lib/types/json.js
```diff
@@ -149,12 +150,12 @@ function json (options) {
...
  */
 
-function createStrictSyntaxError (str, char) {
+function createStrictSyntaxError (parser, str, char) {
   var index = str.indexOf(char)
   var partial = str.substring(0, index) + '#'
 
   try {
-    JSON.parse(partial); /* istanbul ignore next */ throw new SyntaxError('strict violation')
+    parser(partial); /* istanbul ignore next */ throw new SyntaxError('strict violation')
```
dougwilson: This is the line my above comment was supposed to be on.

dougwilson: I'm not sure this is right for custom parsers. Or at least it's not clear in the docs for the parser option that this is going to happen and how people should build their parser function to compensate.

lib/types/json.js
shawndellysse: Actually, I don't think I can agree with you that the docs are not clear: 

> The reviver option is passed directly to the `parser` as the second argument.

If you can suggest an update to that text I'm all ears.

dougwilson: I moved my comment down. The new comment for this line is to clarify in the docs the arguments the custom parser will be called with.

dougwilson: Remove the line from the reviver docs and add in the specific signature into the parser docs. The current parser section doesn't help someone build  their own, I think it assumes someone is going to understand JSON.parse and copy that? If so, just write that down.

dougwilson:
Ok, just wanted to give feedback real quick. I'll be gone for tge holiday and back next week, so see how far you can get on the changes by then or let me know asap if you have comments on the overall request I made instead of just those line comments.

dougwilson:
I love the changes! I'm not sure if you responded to my general comment, maybe I missed it. Why not implement https://github.com/expressjs/body-parser/issues/22 to resolve the underlying issue? I'm just curious because that's still needed and if course that eould replace this causing more API churn in the log run.

README.md
```diff
@@ -86,16 +86,27 @@ specifies the number of bytes; if it is a string, the value is passed to the
 [bytes](https://www.npmjs.com/package/bytes) library for parsing. Defaults
 to `'100kb'`.
 
+##### parser
+
+The `parser` option is the function called against the request body to convert
```
dougwilson: No description of the first argument. Also doesn't describe the mechanism of how it needs to be made to construct the syntax error during the recalling mechanism with the # inserted into the original string.

lib/types/raw.js
```diff
@@ -49,7 +50,7 @@ function raw (options) {
     : type
 
   function parse (buf) {
-    return buf
+    return parser(buf)
```
dougwilson: We probably don't need two levels of functions here. Just replace parse with a variable.

lib/types/text.js
```diff
@@ -51,7 +52,7 @@ function text (options) {
     : type
 
   function parse (buf) {
-    return buf
+    return parser(buf)
```
dougwilson: We probably don't need two levels of functions here. Just replace parse with a variable.

README.md
```diff
@@ -159,6 +170,15 @@ The `verify` option, if supplied, is called as `verify(req, res, buf, encoding)`
 where `buf` is a `Buffer` of the raw request body and `encoding` is the
 encoding of the request. The parsing can be aborted by throwing an error.
 
+##### parser
+
+The `parser` option, if supplied, is used to transform the body of a request
```
dougwilson: Should probably describe the type and number of arguments it will be called with.

README.md
```diff
@@ -208,6 +228,15 @@ The `verify` option, if supplied, is called as `verify(req, res, buf, encoding)`
...
```
dougwilson: Should probably describe the type and number of arguments it will be called with.

README.md
```diff
@@ -274,6 +303,16 @@ The `verify` option, if supplied, is called as `verify(req, res, buf, encoding)`
 where `buf` is a `Buffer` of the raw request body and `encoding` is the
 encoding of the request. The parsing can be aborted by throwing an error.
 
+##### parser
+
+The `parser` option, if supplied, is used to in place of the default parser to
```
dougwilson: Should probably mention the type of the argument and also that it is not called at all for zero length bodies (though, should it be always called because what if they want a zero length result to be something other than {}?). The other parsers seem to be called on zero length, so this seems odd not to.

README.md
```diff
@@ -274,6 +303,16 @@ The `verify` option, if supplied, is called as `verify(req, res, buf, encoding)`
...
+convert the request body into a Javascript object. If this option is supplied,
+both the `extended` and `parameterLimit` options are ignored.
```
dougwilson: For the json parser, the reviver is not ignore but these are. Should json work like this and ignore the arguments that were originally going to the underlying parser that is now replaced, or should these be passed to the new urlencoded parser somehow or neither and a warning is a user gave these? Some thoughts.

shawndellysse:
Closing, superceded by PR #282

PR #283, opened 2017-12-03
Update debug to latest version

DiegoRBaquero:
And get patches


dougwilson:
Thanks so much! I'm aware of the 3.x series but we cannot upgrade our 1.x line to that version if debug as it will break existing user's debug workflows. Can you make the following xhanges?

1. Retarget the 2.0 branch
2. Add an entry to the HISTORY file describing the version bump and the things changed between the previous version and this version.
3. Change the version to be specific instead of a range.

Thanks!

DiegoRBaquero:
Is 2.0 ever gonna come out? It's been stale for years. 263 commits behind master.

Plus, why would it break the current workflow?

dougwilson:
For one *DEBUG_FD* was removed, breaking users who use that to redirect to a named fd im a pass for example. The output is a different format and also timers suddenly have a very different meaning.

dougwilson:
I'll rebase the 2.0 branch for you tomorrow if you want it to be more up-to-date first. All 2.0 was waiting for was the generic body parser, and a PR was just recently made for it, so that will close up the 2.0 release on landing.

DiegoRBaquero:
Alright, yes, I'd prefer. I'm gonna close this PR and make another one from the update 2.x branch once rebased. Thanks!

PR #326, opened 2018-07-31
urlencoded: Support iso-8859-1, utf8 sentinel, and numeric entities

### LINKED ISSUE #194: iso-8859-1/windows-1252 support?
Hi,

A while ago I developed a web app to replace an old cgi-bin script from the nineties, which means that it had to assume that POST requests with a `Content-Type` of `application/x-www-urlencoded` are in `iso-8859-1` (and also support "smart quotes" etc., aka. windows-1252). At the same time it had to be possible to switch to `utf-8` mode.

I accomplished that by adding support for a `defaultCharset` option, which can either be `utf-8` or `iso-8859-1`, and for explicitly switching to `utf-8` mode by adding `utf8=%E2%9C%93` (`utf8=✓` as percent-encoded utf-8 octets) to the query string, a trick that I've seen others use as well.

Here are the diffs between my forks of body-parser and the qs module:

https://github.com/expressjs/body-parser/compare/master...papandreou:iso-8859-1
https://github.com/ljharb/qs/compare/master...papandreou:interpretNumericEntities

Would anyone be interested in this work? If so, I can clean it up and open PRs.

papandreou:
Fixes #194

Further elaborations here: https://github.com/ljharb/qs/pull/268

Supports both the simple and extended parsers as outlined in https://github.com/ljharb/qs/pull/268#issuecomment-409195381 -- and doesn't use the equivalent capabilities of the qs module.

~Note: This will fail in node.js 0.8 and 0.10 as the capability to specify a custom `decodeURIComponent` implementation to use wasn't added until 0.12.~


dougwilson:
> Note: This will fail in node.js 0.8 and 0.10 as the capability to specify a custom decodeURIComponent implementation to use wasn't added until 0.12.

We can hold the PR for the 2.0 release, which will drop support for those versions 👍

papandreou:
@dougwilson, that sounds fair! Does it look good otherwise?

Lurow:
Hi there,

Any news on v2.0 release date with the pr?

Cheers

thopaw:
Any news to iso-8859-1 support?

papandreou:
Fixed conflict with master.

@dougwilson, this has been sitting for some time now :sweat_smile:

I'm not really in a hurry, but how about getting this released soon? I can remove those ancient node versions from `.travis.yml` if that would be a help? (I guess it might be a nice occasion to drop more old versions)

dougwilson:
The reason it has been sitting is just because it cannot be merged into the 1.x line due to the incompatibility. Because this module is a part of express, it effectively inherits it's support policy. The 1.x series of this module won't end up out of support for some time, and the count down starts with the next express release.

In order to reduce the support burden by maintaining multiple major versions of this module, I am just waiting to make the 2.0 in coordination with express 5.

If this could be made to work in those node.js versions I would of course be happy to land right away.

papandreou:
Okay, thanks for the update, that makes sense!

> If this could be made to work in those node.js versions I would of course be happy to land right away.

Unfortunately I think this would involve us to stop relying on the built-in `querystring` module in simple mode. Either by copying in a newer implementation than the 0.10, or by figuring out a way to use the `qs` module in simple mode as well, disabling the advanced behaviors. That seems a bit far-reaching, unless you were already planning to go in that direction?

ljharb:
@papandreou if there's manageable changes i could make in `qs` that would allow body-parser to switch to it for everything, i'm willing to make those.

dougwilson:
And I would be happy to use `qs` for everything here as well. The downside to using `querystring` for the simple parser is that since it ships as part of Node.js core, the results of this module will vary by both the version of this module and the version of Node.js (vs only varying by the version of this module).

ljharb:
@dougwilson if you can save me a few minutes and link me to a branch that uses qs for the simple parser, but is failing tests, i can use npm link and see what options and changes it would take to get it working :-D 🙏

dougwilson:
@ljharb this is likely not even correct for params to `qs` (I passed `depth: -1` since `0` is not the right behavior, but idk), but it's at least _something_ that should help :)

Branch: https://github.com/expressjs/body-parser/tree/urlencoded-qs-simple
Test: `npm i && npm test -- --grep=bodyParser.urlencoded`

ljharb:
give me 48 hours, i'll see what i can do

ljharb:
body-parser tests on that branch now pass with https://github.com/ljharb/qs/pull/326#issuecomment-522157740, and I can merge and release that in `qs` as soon as I've convinced myself what semver it is.

dougwilson:
Ah, yes, so depth: 0 was supposed to make sense after all :D . Yes, that is a hard decision. But if it helps at all, this module and Express do not directly expose the qs arguments, making whichever semver it is a non issue to us :)

ljharb:
What I'm probably leaning towards is releasing it as a minor, and if users report breakage, reverting the depth 0 part, but keeping the depth false part - so express thus should use depth false to ensure continuity.

ljharb:
v2.8.0 of qs is published.

filecage:
The PR seems to be ready and all dependencies resolved. Is there anything left blocking this PR to be merged (except for the `package.json` conflict) and released as minor that I can help with?

Even in 2020 it is still interesting to have ISO-8859-1 support in web apps. 😬

ljharb:
might be good to make sure it's using the latest qs, v6.9.4, but otherwise i'd still love to see this land as well.

papandreou:
Resolved the conflict and updated to 6.9.4 :innocent:

DarkBitz:
Is there a specific reason why the pull request is not being merged?

wesleytodd:
I merged #387 yesterday, I am trying to catch up on what was going on here but was thinking it might be a better idea to see if you (@papandreou) wanted to get this PR updated with that in mind? Or are we calling this not necessary anymore?

ljharb:
> Note: This will fail in node.js 0.8 and 0.10 as the capability to specify a custom decodeURIComponent implementation to use wasn't added until 0.12.

means it should probably wait for a major?

wesleytodd:
A major is what I am doing! So that is why I am circling back on all these PRs.

ljharb:
in that case it seems like a great idea

wesleytodd:
With that in mind, I am not up on the specifications involved here and am just doing the mechanical turk job of preparing all these stale branches for prime time, can you give a review and TLDR of what we should be aware of with this and #387 going out along with `express@5`?

README.md
```diff
@@ -277,6 +275,23 @@ The `verify` option, if supplied, is called as `verify(req, res, buf, encoding)`
 where `buf` is a `Buffer` of the raw request body and `encoding` is the
 encoding of the request. The parsing can be aborted by throwing an error.
 
+##### defaultCharset
+
+The default charset to parse as, if not specified in content-type. Must be
+either `utf-8` or `iso-8859-1`.
```
ljharb, line 281: this allows changing the charset, which is a nice semver-minor; the default should probably be utf-8

README.md
```diff
@@ -277,6 +275,23 @@ The `verify` option, if supplied, is called as `verify(req, res, buf, encoding)`
...
+
+##### charsetSentinel
+
+Whether to let the value of the `utf8` parameter take precedence as the charset
+selector. It requires the form to contain a parameter named `utf8` with a value
+of `✓`. Defaults to `false`.
```
ljharb, line 287: this goes along with `defaultCharset`

README.md
```diff
@@ -277,6 +275,23 @@ The `verify` option, if supplied, is called as `verify(req, res, buf, encoding)`
...
+
+##### interpretNumericEntities
+
+Whether to decode numeric entities such as `&#9786;` when parsing an iso-8859-1
+form. Defaults to `false`.
```
ljharb, line 292: and this goes along with the other defaultCharset value

lib/types/urlencoded.js
```diff
@@ -19,18 +19,76 @@ var debug = require('debug')('body-parser:urlencoded')
...
+  // application/x-www-form-urlencoded body and the encoding of the page containing
+  // the form is iso-8859-1, or when the submitted form has an accept-charset
+  // attribute of iso-8859-1. Presumably also with other charsets that do not contain
+  // the ✓ character, such as us-ascii.
+  '%26%2310003%3B': 'iso-8859-1', // encodeURIComponent('&#10003;')
+  // These are the percent-encoded utf-8 octets representing a checkmark, indicating
+  // that the request actually is utf-8 encoded.
+  '%E2%9C%93': 'utf-8' // encodeURIComponent('✓')
+}
+
```
ljharb, line 41: this does confuse me a bit tho; it kind of seems like code that already exists in qs? meaning that the options can just be passed to qs and it'll handle the logic.

altho if someone's using a different querystring parser then the logic might need replicating

ljharb (commented):
Reviewed some of it, but i'm not familiar enough with body-parser to review the rest.

papandreou:
> I merged #387 yesterday, I am trying to catch up on what was going on here but was thinking it might be a better idea to see if you (@papandreou) wanted to get this PR updated with that in mind? Or are we calling this not necessary anymore?

I think it's still a nice-to-have! I've merged the `2.x` branch in and resolved the conflicts now.

lib/types/urlencoded.js
papandreou: You're correct that these magic values are also found [here](https://github.com/ljharb/qs/blob/f90cc35dd65c7099c35ae75d7a1a67aab85220e1/lib/parse.js#L49-L52). I spent some time trying to remember why also having the decoding logic here was warranted, but it seems like [we can get actually get rid of it](https://github.com/papandreou/body-parser/commit/cbf4cec5c601473c25226d205fcaa84c57706308) and still have the test suite pass!

I wonder if I did it like this because it seemed like a long shot to get changes into the `qs` module, so I went for `body-parser` first, because that's what I needed myself, if I recall correctly. It has been 6 years, though, so I'm not really sure 🤔

README.md
papandreou: It is. Documented that in 44a6aa944c05cd865cb2479e2e2d00c816c02c76 now.

ljharb (commented):
LGTM - altho i don’t actually see how this is breaking? where is a custom decodeURIComponent implementation used?

papandreou:
> altho i don’t actually see how this is breaking? where is a custom decodeURIComponent implementation used?

Ah, that referred to the `querystring` module built into node.js. In node.js 0.12 and above, it supports specifying a custom `decodeURIComponent` implementation ([docs](https://nodejs.org/docs/latest-v0.12.x/api/querystring.html#querystring_querystring_parse_str_sep_eq_options)), whereas 0.10 and below didn't ([docs](https://nodejs.org/docs/latest-v0.10.x/api/querystring.html#querystring_querystring_parse_str_sep_eq_options)).

Before expressjs/body-parser#387, `body-parser` would use the `querystring` module when not configured to use the `extended` parsing mode. I _think_ that meant that without expressjs/body-parser#387, the new features introduced in this PR would only have worked in `extended` mode in node.js 0.10 and below. Even 6 years ago, node.js 0.12 had been out for 6 years, and it seemed fine to just live with it and do a new major release. And here we (almost) are 😆

wesleytodd:
Well either way I don't think we are doing any extra work to land these things in the current majors. We have needed to clean house on much of this for a while, so as long as this is good to land we will be doing it along with a major no matter what. And with that in mind, here I am clicking the merge button. We will be preparing the v2 release soon.

PR #338, opened 2018-11-17
Add String or Number Only JSONString Support

itorr:
```javascript
JSON.parse(`"text"`) // "text"
JSON.parse(`123`) // 123
```


dougwilson:
https://github.com/expressjs/body-parser/blob/master/README.md#strict

PR #362, opened 2019-06-04
Changed Examples To ES6

KengoWada:
I have changed the code samples in the README.md to follow the es6 standard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,9 +396,6 @@ const jsonParser = bodyParser.json();
 
 // create application/x-www-form-urlencoded parser
 const urlencodedParser = bodyParser.urlencoded({ extended: false });
-var
-var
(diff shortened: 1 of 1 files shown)


wesleytodd:
Hi @KengoWada, thank you for the submission.  Unfortunately this is probably not something we will merge as is.  There are a few reasons:

1. There are unrelated changes to the markdown style (`-` => `*` for example)
2. The formatting of the examples was changed.  The Express project uses `standard` js style, and the new example format does not follow this

If the PR just had the `var` => `const` changes I think we would probably accept it.  If you are up for making these changes you might as well also squash it down to a single commit.

Thanks!

KengoWada:
Thanks for the feedback.

PR #363, opened 2019-06-04
Changed examples to es6

KengoWada:


dougwilson:
Thank you, but the idea for the example syntax is for it to work on all the Node.js versions that the module itself supports.

PR #374, opened 2019-09-26
Support for checking empty JSON body.

chainhead:
# Requirement
Let the JSON body parser package raise an error if the payload is empty. This middleware maybe added for certain `POST` or `PUT` requests.

## Solution
Check for an empty string and raise a new error named as `createEmptyBodyError` which is similar to `function createStrictSyntaxError` in `parse` function at [`lib/tpyes/json.js`](https://github.com/expressjs/body-parser/blob/b15bae549108e613ceb5e408731cbdb374ad62df/lib/types/json.js#L71). This is done when the option of strict checking is set to `true` - the default.

## Notes

1. When the body is empty, then `length` is not set correctly at [`lib/read.js`](https://github.com/expressjs/body-parser/blob/b15bae549108e613ceb5e408731cbdb374ad62df/lib/read.js#L146), if `Content-Length` is not passed in the request. However, adding a check for `Content-Length` header may conflict prior installations.*
2. For requests that rely on a payload, a 'strict' setting **should** mean a valid and a non-empty JSON payload. In other words, an extra `options` value to test for empty/non-empty payloads may not be required.

* - For my purpose, I have a middleware before JSON body parser that looks for multiple headers including `Content-Length` for this requirment.

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -47,7 +47,7 @@ var FIRST_CHAR_REGEXP = /^[\x20\x09\x0a\x0d]*(.)/ // eslint-disable-line no-cont
  * @public
  */
 
-function json (options) {
+function json(options) {
   var opts = options || {}
 
   var limit = typeof opts.limit !== 'number'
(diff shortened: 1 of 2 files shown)


wesleytodd:
Hi @chainhead, can you write more about your use case?  Is there a reason this module should be doing this, and not some other middleware specifically for the purpose?

One other small note, this PR also includes unrelated style changes, making it more difficult to tell what is being changed.  Can you remove those?

lib/types/json.js
```diff
@@ -104,12 +108,15 @@ function json (options) {
 
     req.body = req.body || {}
 
+    // BEGIN - Check empty body
+    /*
```
dougwilson: I believe this has been left commented out on accident -- this is how all the parsers work, they only parse bodies and this does not parse when there is no body at all (this is different from a zero-length body, i.e. empty string, which AFAIK is that you are adding an error for).

lib/types/json.js
```diff
@@ -150,15 +157,34 @@ function json (options) {
...
+function createStrictSyntaxError(str, char) {
   var index = str.indexOf(char)
   var partial = str.substring(0, index) + '#'
 
   try {
     JSON.parse(partial); /* istanbul ignore next */ throw new SyntaxError('strict violation')
   } catch (e) {
     return normalizeJsonSyntaxError(e, {
-      message: e.message.replace('#', char),
+      message: e.message.replace('#', char) + ' NS ',
```
dougwilson: I'm not sure I understand this message here. `NS`?

wesleytodd:
I am going to close this. If we would like to revisit this in the future please open a new updated PR with a description about why we should add this.

PR #378, opened 2019-10-08
Expands Errors and adds 'entity.parse.failed'

HopefulLlama:
Expands on errors:
* States that they are the originally thrown error, with additional properties
* Adds `entity.parse.failed`

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,6 +300,13 @@ contained an encoding but the "inflation" option was set to `false`. The
 `'encoding.unsupported'`, and the `charset` property will be set to the
 encoding that is unsupported.
 
+### entity parse failed
+
+This error will occur when the request contained an entity that could not be
+parsed by the middleware. The `status` property is set to `400`, the `type`
+property is set to `'entity.parse.failed'`, and the `body` property is set to
+the entity value that failed parsing.
+
 ### request aborted
 
 This error will occur when the request is aborted by the client before reading


README.md
```diff
@@ -282,11 +282,22 @@ encoding of the request. The parsing can be aborted by throwing an error.
...
 
 The middlewares provided by this module create errors depending on the error
-condition during parsing. The errors will typically have a `status`/`statusCode`
-property that contains the suggested HTTP response code, an `expose` property
-to determine if the `message` property should be displayed to the client, a
-`type` property to determine the type of error without matching against the
-`message`, and a `body` property containing the read body, if available.
+condition during parsing. 
+
+This is passed to an Express error handling middleware, if one is defined,
```
dougwilson: Is this necessary to say? This is how all express middleware work and it would be nice if I didn't need to add this paragraph to every middleware module's readme. Is there is reason that would lead one to believe that errors from middleware are not typically passed to express error handling middleware? This module is usable outside express as well, so if this is to be kept, it likely needs to better explain how the errors are passed for the non express users (right now it is omitted as one would refer to their frameworks docs on handling middleware errors).

README.md
```diff
@@ -282,11 +282,22 @@ encoding of the request. The parsing can be aborted by throwing an error.
...
-to determine if the `message` property should be displayed to the client, a
-`type` property to determine the type of error without matching against the
-`message`, and a `body` property containing the read body, if available.
+condition during parsing. 
+
+This is passed to an Express error handling middleware, if one is defined, 
+in the usual way. Therefore, further inspection and handling can be performed 
+in another error handling middleware.
+
+The error provided is the original error thrown, with additional properties to
```
dougwilson: I'm not sure if this part actually clarifies anything. Like how would one know what "the original error thrown" is? Especially since most of the errors here are ever actually thrown at all internally?

README.md
HopefulLlama: My reasoning behind including this, is that body-parser is one of the more popular packages related to Express, and it's therefore possible that body-parser users aren't Express experts who are looking to configure error handling, and mistakenly think that body-parser is the place to configure it. 

Perhaps it'd be more useful to add a link to the original Express documentation regarding error handling middleware?

This way it'd retain its purpose for non-expert users who are looking here for information, while keeping the onus of documenting error handling on Express.

README.md
HopefulLlama: My intention here was to clarify that the error passed is isn't simply the additional properties, but also the actual Error object which contains name, message, stack, etc properties.

Unsure on how to phrase it better, but welcome to any suggestions or edits you have in mind.

README.md
dougwilson: I don't think Express users will end up here as the primary method to parse bodies, as express has this built into itself, as express.json, express.urlencoded, etc.

README.md
dougwilson: At least state that it is an Error object somewhere, otherwise it doesn't really say anything different than it did before, as it also said these were additional properties, not the only properties, prior to this change.

README.md
HopefulLlama: I take your point. Happy to remove this section in thiscase.

PR #385, opened 2019-12-31
fix(deps): added nodejs 13 and all operating system for travis build

smondal:
Added travis build against all operating system and updated nodejs version 13.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+os:
+  - linux
+  - osx
+  - windows
 language: node_js
 node_js:
   - "0.8"
@@ -15,6 +19,7 @@ node_js:
   - "10.16"
   - "11.15"
   - "12.4"
+  - "13.5"
 sudo: false
 cache:
   directories:


.travis.yml
```diff
@@ -1,3 +1,7 @@
+os:
+  - linux
+  - osx
+  - windows
```
dougwilson: The expressjs project currently only tests against different operating systems if there is some kind of os-dependent code path like file operations or path manipulation. Does this apply to this project, and if so, please elaborate on how :+1:

smondal: See the travis log its breaking against windows operating system. 
Here is the logfile. 

Downloading bootstrap node from https://nodejs.org/dist/v10.12.0/node-v10.12.0-win-x64.7z
######################################################################## 100.0%
The command "nvs add 1.8" failed and exited with 1 during .

dougwilson: The expressjs project tests windows using AppVeyor and not using Travis CI.

dougwilson:
Closing due to no response / CI failing.

PR #414, opened 2020-10-01
Make it work with Async Hooks

kjarmicki:
This is the reimplementation of #404 using `AsyncResource` instead of promise wrappers, as @Qard advised.
Tests were taken in their entirety from the original PR.

### DIFF
--- a/lib/read.js
+++ b/lib/read.js
@@ -17,6 +17,12 @@ var iconv = require('iconv-lite')
 var onFinished = require('on-finished')
 var zlib = require('zlib')
 
+var asyncHooks
+try {
+  asyncHooks = require('async_hooks')
+} catch (ignored) {
+}
+
 /**
  * Module exports.
  */
@@ -74,7 +80,7 @@ function read (req, res, next, parse, debug, options) {
 
   // read body
   debug('read body')
-  getBody(stream, opts, function (error, body) {
+  getBody(stream, opts, preserveAsyncContext('GetBody', function (error, body) {
     if (error) {
       var _error
 
@@ -91,9 +97,9 @@ function read (req, res, next, parse, debug, options) {
 
       // read off entire request
       stream.resume()
-      onFinished(req, function onfinished () {
+      onFinished(req, preserveAsyncContext('OnFinished', function onfinished () {
         next(createError(400, _error))
-      })
+      }))
       return
     }
 
@@ -128,7 +134,7 @@ function read (req, res, next, parse, debug, options) {
     }
 
     next()
-  })
+  }))
 }
 
 /**
@@ -178,4 +184,15 @@ function contentstream (req, debug, inflate) {
   }
 
   return stream
+}
+
+function preserveAsyncContext (asyncResourceType, fn) {
+  if (!asyncHooks) {
+    return fn
+  }
+  var asyncResource = new asyncHooks.AsyncResource(asyncResourceType)
+  return function () {
+    var args = [].slice.call(arguments)
+    asyncResource.runInAsyncScope.apply(asyncResource, [fn, null].concat(args))
+  }
 }
--- a/lib/read.js
+++ b/lib/read.js
@@ -191,8 +191,5 @@ function preserveAsyncContext (asyncResourceType, fn) {
     return fn
   }
   var asyncResource = new asyncHooks.AsyncResource(asyncResourceType)
-  return function () {
-    var args = [].slice.call(arguments)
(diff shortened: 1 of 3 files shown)


lib/read.js
```diff
@@ -179,3 +185,14 @@ function contentstream (req, debug, inflate) {
...
 }
+
+function preserveAsyncContext (asyncResourceType, fn) {
+  if (!asyncHooks) {
+    return fn
+  }
+  var asyncResource = new asyncHooks.AsyncResource(asyncResourceType)
+  return function () {
+    var args = [].slice.call(arguments)
+    asyncResource.runInAsyncScope.apply(asyncResource, [fn, null].concat(args))
```
Qard, line 196: Rather than this apply here, you could replace the whole function this is wrapped in by using bind. Something like return asyncResource.runInAsyncScope.bind(asyncResource, fn)

kjarmicki: Oh, that's way nicer :slightly_smiling_face:. Thanks.

lib/read.js
```diff
@@ -74,7 +80,7 @@ function read (req, res, next, parse, debug, options) {
 
   // read body
   debug('read body')
-  getBody(stream, opts, function (error, body) {
+  getBody(stream, opts, preserveAsyncContext('GetBody', function (error, body) {
```
dougwilson, line 83: Should not the `getBody` function itself be preserving the async context? Wouldn't adding it within `getBody` function directly preserve the context for every dependency on npm that uses it instead of needing to work through every one of them (like this module) to add this? It seems quite burdensome for all the dependencies of that module if this is is true. Because by that logic, why don't you do this around the call to `bodyParser` itself?

lib/read.js
```diff
@@ -91,9 +97,9 @@ function read (req, res, next, parse, debug, options) {
 
       // read off entire request
       stream.resume()
-      onFinished(req, function onfinished () {
+      onFinished(req, preserveAsyncContext('OnFinished', function onfinished () {
```
dougwilson, line 100: Should not the `onFinished` function itself be preserving the async context? Wouldn't adding it within `onFinished` function directly preserve the context for every dependency on npm that uses it instead of needing to work through every one of them (like this module) to add this? It seems quite burdensome for all the dependencies of that module if this is is true. Because by that logic, why don't you do this around the call to `bodyParser` itself?

dougwilson (changes requested):

lib/read.js
kjarmicki: That's true, in principle this is the right thing to do, and also that's the reason why the discussion on how to handle async context in event handlers is happening now between NodeJS folks, as it would be the lowest layer that benefits all the dependent modules.

In practice though, having contributed to third party modules in the past, I found it quite difficult to get the attention of the maintainers to even look at the PR, let alone merging it. So I figured `body-parser` to be my best shot, as it was a single direct dependency of my apps that was causing issues (rather than two transitive ones)  and it looked alive on GitHub.

Having said that, I see that you're a collaborator on both of these, so I presume that's not the problem here :slightly_smiling_face: If you're willing to pull in my PRs there, release new versions, let me update their declarations in `body-parser` and then release it, I'm happy to do it and we can close this PR.

lib/read.js
Qard: Yep, adding it directly to getBody is probably better.

lib/read.js
Qard: Yep, same as getBody, I'd recommend this be added to onFinished.

edingroot:
May I know why this PR is closed?

edingroot:
May I know why this PR was closed?

kjarmicki:
@edingroot `body-parser` is using `raw-body` and `on-finished` which are the lowest level modules that could have the fix applied and proliferate upwards from there, so it made sense to introduce it at the level of these modules.
I'm happy to say that there PRs were already merged 🙂 see https://github.com/stream-utils/raw-body/pull/71 and https://github.com/jshttp/on-finished/pull/36

I'm not sure when there's going to be an Express update that transitively pulls these changes in, but if you're in a hurry I think you could tackle it with npm overrides or patch-package.

dougwilson:
I am working on integrating them in to the very next body-parser release. There was a reported potential security issue with body-parser I am investigating is the only hold up. I pushed up the necessary changes, but rolled them back in anticipation of a patch release if this turns out to indeed be a security issue.

Everything should be done this week.

edingroot:
Got it! Thank you all for the great work :)

dougwilson:
The changes are staged on `master` now if anyone wants to test, but will be released in the coming days :) !

PR #417, opened 2020-11-19
Update debug @4.x

ecasilla:


dougwilson:
Thank you, unfortunately this is not compatible with the versions on Node.js supported by 1.x. Would you be willing to make this PR against the 2.0 branch?

ecasilla:
Actually for 2.x branch there seems to be no reference to the debug module so maybe its no needed I checked both the 2.x and 2.x-refactor branches. Are these branches available for use or is their a timeline for there release?

PR #423, opened 2021-01-31
Upgrade deps to latest

lopopolo:
Add range version specifiers.

If this patch is accepted and merged, can the maintainers push a new release to npm?


dougwilson:
Many of these major version bumps cause the CI to fail. I suggest perhaps splitting the PR to one per dep so they can be evaluated one by one.

In addition, we don't accept third party dependencies to be ranges.

PR #425, opened 2021-02-13
Updated README as per ES6 conventions.

Deveshb15:
Replaced var keyword with const, wherever applicable as per ES6 conventions


dougwilson:
Thank you, but this is a duplicate of #363

PR #444, opened 2021-08-07
Updated Debug library

sandeshvijayraj:
Debug Library < 4.1.1 are detected as vulnerable and may be blockers to many projects having vulnerability scan enabled.

This PR updates Debug Library to the latest version to avoid the library getting detected as vulnerable

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "dependencies": {
     "bytes": "3.1.0",
     "content-type": "~1.0.4",
-    "debug": "2.6.9",
+    "debug": "^4.3.2",
     "depd": "~1.1.2",
     "http-errors": "1.7.3",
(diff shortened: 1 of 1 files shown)


dougwilson:
The issue is upgrading this is a breaking change. The plan for it js part of the 2.0 plan, though. As for it being vulnerable, there is no vulnerability that affects this specific version I am aware of, as the author published out various major releases with security fixes and we are on the most recent 2.x release. I checked out a few of the public npm vulnerability databases and none of them mark this version of debug as vulnerable to anything, either.

dougwilson:
I just double checked the debug project itself and there was no security release in 4.1.1 and the only recent security issue that happened after this last 2.x release only affected select versions of 3.x and 4.x (as it was a regression from a change committed there).

PR #457, opened 2022-02-05
chore(std bump): Updated debug's version to 4.3.3

ihack2712:
There's an issue with debug showing different timestamp formats on different versions. As far as I'm aware it introduces no errors to upgrade.

debug-js/debug#867

@Qix- asked to be mentioned

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "dependencies": {
     "bytes": "3.1.2",
     "content-type": "~1.0.4",
-    "debug": "2.6.9",
+    "debug": "4.3.3",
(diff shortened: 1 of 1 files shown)


dougwilson:
Hello, and thank you for your pull request. As you can probably see from the CI failures, the new version of debug you used here is not compatible as the same set of Node.js versions as body-parser 1.x . The debug library is already being upgraded in the next semver major of body-parser, though.

iddan:
Hey @dougwilson where can I follow the progress of the next semver major of body-parser?

Issue #100, opened 2015-05-11
Content-Encoding issue


cha55son:
I've got an application that is sending webhooks to hubot. I receive the error pasted below. Unfortunately the app is sending webhooks with a header of **Content-Encoding: UTF-8** and there is nothing I can do on that end. I can change hubot however I need to though. Is there a way to get express/body-parser to "ignore" the Content-Encoding header?

This used to work with an older version of hubot. I just recently upgraded and i'm now hitting this error. Any help or direction would be appreciated.

```
Error: unsupported content encoding "utf-8"
  at contentstream (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/node_modules/body-parser/lib/read.js:146:13)
  at read (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/node_modules/body-parser/lib/read.js:42:14)
  at Object.jsonParser [as handle] (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/node_modules/body-parser/lib/types/json.js:96:5)
  at next (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/lib/proto.js:169:15)
  at Object.query [as handle] (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/lib/middleware/query.js:43:5)
  at next (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/lib/proto.js:169:15)
  at Object.handle (/home/cchoate/drbot/node_modules/hubot/src/robot.coffee:288:7, <js>:243:16)
  at next (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/lib/proto.js:169:15)
  at Object.expressInit [as handle] (/home/cchoate/drbot/node_modules/hubot/node_modules/express/lib/middleware.js:30:5)
  at next (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/lib/proto.js:169:15)
  at Object.query [as handle] (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/lib/middleware/query.js:43:5)
  at next (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/lib/proto.js:169:15)
  at Function.app.handle (/home/cchoate/drbot/node_modules/hubot/node_modules/express/node_modules/connect/lib/proto.js:177:3)
[...]

PR #469, opened 2022-09-01
Fix error content-encoding type

AnsonZnl:
fix this issues: https://github.com/expressjs/body-parser/issues/100


dougwilson:
Thank you for your pull request, but this is not a change that we are planning to accept. You can read more in the issue you linked.

PR #472, opened 2022-11-10
Add post-parsing verify function

jhford-scout24:
Allow user to specify a function to run on the parsed output.  This function gets the req and res as does the normal parsing middlewares but operates on the parsed data.

This patch is only a sketch to illustrate my idea, I haven't implemented tests or anything, but wanted to see if such a feature is of interest to the project.  I would also add this functionality to the other parsed body types if the patch is acceptable.

My specific use case is that i want to use bodyparser with json-schema. I wanted to use the verify function option, but in order to do that I would need to reparse the json.  I could also wrap the bodyparser middleware in my own middleware, but this is more complicated than adding for example `verifyParsed: (rq,rs,b) => {validate(b)}` to the bodyparser middleware.

### DIFF
--- a/lib/read.js
+++ b/lib/read.js
@@ -50,6 +50,7 @@ function read (req, res, next, parse, debug, options) {
     ? opts.encoding
     : null
   var verify = opts.verify
+  var verifyParsed = opts.verifyParsed
 
   try {
     // get the content stream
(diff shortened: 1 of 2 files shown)


wesleytodd:
Hey @jhford-scout24, sorry for the very much delayed response. The express project has struggled with investment for some years now but we are trying to get back on track. I just wanted to drop a note in here though to see if you were still interested in having this conversation. I have some thoughts, but we need to ruthlessly prioritize the required work to get the next majro released, and it is highly unlikely we would land *this* PR for that. That said, I think I am interested in thinking about this problem space a bit more clearly, and I am interested in `express` having this type of functionality even if it does not live inside body parser itself. If that is not how you see it being used though I would like to hear that.

PR #475, opened 2022-11-29
chore: fix tests for node 19

nlf:
there was a slight change in the error messaging for JSON.parse that caused some assertions to fail so i added a second parameter to alter the input we're asserting against to more appropriately match the expected error. this gets tests passing in node 19+

### DIFF
--- a/test/json.js
+++ b/test/json.js
@@ -245,7 +245,7 @@ describe('bodyParser.json()', function () {
           .post('/')
           .set('Content-Type', 'application/json')
           .send('true')
-          .expect(400, '[entity.parse.failed] ' + parseError('#rue').replace('#', 't'), done)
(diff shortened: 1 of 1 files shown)


dougwilson:
Hi @nlf thanks for the reminder! The change is more than just a test issue, as it changes the actually way the error messages are constructed too (you'll notice the output when using the module is incorrect under Node.js 19). I started the fix on a branch with commit https://github.com/expressjs/body-parser/commit/495ee645a4e55f131df26bab7fb59aac21dd32b4 . I was testing it and it seems fine, just forgot to get it past the finish line.

nlf:
no worries! you're right, your commit looks good and seems to pass tests for me. this can be closed at any time. thanks @dougwilson!

PR #485, opened 2023-01-11
deps: debug@3.1.0

ashkulz:
fix for CVE-2017-20165 -- CI [passes on my fork](https://github.com/ashkulz/body-parser/actions/runs/3889489280).


ashkulz:
@dougwilson would appreciate merging this and pushing a new release to npm, as this is causing vulnerability reports in our repository.

dougwilson:
Hello, and thank you for your pull request. Though your change passes our tests, the major version upgrade of `debug` will change how users interact with the `debug` part of this module in a breaking way. This change is already available in `body-parser` 2.0.

I took a look at your link and under references it links to that Github PR https://github.com/debug-js/debug/pull/504 as the fix which, if you look there, it shows that the project also put that fix as debug 2.6.9, which is already what `body-parser` 1.x is using. It would seem that the version range on the CVE you are referencing is just incomplete and needs to be updated to include the fact that 2.6.9 is also a fix release for debug 2.x just as 3.1.0 was a fix release for debug 3.x.

dougwilson:
> @dougwilson would appreciate merging this and pushing a new release to npm, as this is causing vulnerability reports in our repository.

It looks like we both commented at the same time, so I didn't see this new comment when I authored my first reply. If you are using an automated scanning tool that is incorrectly reporting 2.9.6 as vulnerable, you will need to report this to the tool you are using.

ashkulz:
Thanks for the quick response!

PR #491, opened 2023-02-09
deps: debug@3.1.0

NataliaRequejoSouthworks:
Demo library version 2.6.9 has vulnerability issues and we should update it to at least version 3.1.0
Here is the link for more information: https://vulners.com/osv/OSV:GHSA-9VVW-CC9W-F27H
image

All tests passed without issues
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "dependencies": {
     "bytes": "3.1.2",
     "content-type": "~1.0.5",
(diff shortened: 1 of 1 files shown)


dougwilson:
Hello, and thann you for your pull request. As you can see in the link you provided, 2.6.9 is not vulnerable.

Issue #478, opened 2022-12-18
support for ndjson


pinkasey:
It would be nice if the `json` middleware could support [ndjson](http://ndjson.org/) as well.
e.g, something like this as input:
```
{ "id": 1, "message": "hello" }
{ "id": 2, "message": "I'm a JSON inside a NDJSON" }
```

I've recently tried to consume a webhook of some big SaaS company, and they call the webhook with a POST and `Content-Type: application/json` (they should be passing `Content-Type: application/x-ndjson`).  

From my perspective -
I didn't even know that ndjson exist until yesterday, and it's so similar to json - I would've hoped anything that can consume json can also consume ndjson. It's very easy to make that work.

What I had in mind -
add an option, something like `allow_ndjson` (default to `false` for backward compatibility), and if it's `true` -
when parsing a JSON fails, try splitting the body by `\n`, and parsing each part as json. return an array of the resulting jsons.
So the example above would result it:
```
[
  { "id": 1, "message": "hello" },
  { "id": 2, "message": "I'm a JSON inside a NDJSON" }
]
```

I'll be happy to contribute PR if there's a chance it would be approved.

PR #513, opened 2024-02-09
Custom JSON parser

vbelius:
To solve the need for custom JSON parses, like in issue [#278] and issue [#478], I've added an option for a custom JSON parser. This is a simpler implementation with more limited scope than the stale [#282] PR.

I was a bit unsure on how to best handle the custom error handling/normalization built into the existing JSON parser. I opted for skipping this when supplied with a custom parser, to allow custom parsers to throw their own errors without manipulation, mostly since the normalizeJsonSyntaxError method removes all keys except message and stack, which might be set on custom and errors and might be useful for those using custom parsers. 

```JS
      // Only normalize errors if using default parser
      // Custom parsers might throw custom errors
      if (parser === JSON.parse) {
        throw normalizeJsonSyntaxError(e, {
          message: e.message,
          stack: e.stack
        })
      } else {
        throw e
      }
```

What do you think @dougwilson? Do you have any other feedback that might help get this merged? Thank you.

PS. My personal motivation to implement this feature is to enable the use of json-bigint for parsing json.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,15 +92,15 @@ to `'100kb'`.
 ##### parser
 
 The `parser` option can be used to supply a custom JSON parser used to 
-convert the request body to a Javascript object. If a `reviver` is 
-supplied, it is supplied as the second argument to this function.
+convert the request body to a Javascript object. If the `reviver` option 
+is supplied, it is used as the second argument to this function.
 
 Defaults to `JSON.parse`.
 
 ##### reviver
 
-The `reviver` option is passed directly to `JSON.parse` as the second
-argument. You can find more information on this argument
+The `reviver` option is passed directly to `JSON.parse` (or custom parser) 
+as the second argument. You can find more information on this argument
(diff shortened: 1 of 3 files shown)


vbelius:
@dougwilson Could I argue that since it's been 7 years without any progress on this issue, and 10 years without any progress on #22, wouldn't an intermediary solution be more beneficial to no solution? It's a very simple PR, not adding lots of code or complexity to the project. It could also be argued for us casual JSON users with, special needs, that it would easier to provide a custom option rather than implement a full custom parser?

I hope you find my contemplations in good spirit, thank you for all your awesome OSS work :)

dougwilson:
Hello, I would argue perhaps if you like to help, please help make the generic parser... This is adding complexity to the JSON parser that is unnecessary, just think how you are adding the different code paths to avoid the JSON.parse error munging and making the json parser now work two different ways.

I can explain the gap in development, but it is not relevant here and unrelated to trying to add a feature. Please if you are going to contribute to a project, it would be nice to respect the design decisions of the project.

PR #517, opened 2024-02-29
Update name as it is already taken

emilcondrea:
The repo backed by existing package does not seem to exist: https://www.npmjs.com/package//body-parser-bigint


emilcondrea:
Please ignore, opened PR to wrong repo

PR #520, opened 2024-03-28
Improved JSON parsing to handle tabs within strings.

oukhrib:
This pull request addresses an issue where the parser encountered errors when handling JSON strings containing tabs (actual tab characters). A test case is included to demonstrate the issue:

``` JSON
{"fullname":"John	Doe"}
```
To ensure wider compatibility, the fix replaces all tabs within strings with two spaces. This approach aligns with common parsing practices and avoids potential errors caused by tabs.

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -86,6 +86,11 @@ function json (options) {
         throw createStrictSyntaxError(body, first)
       }
     }
+    // Special case : if the body contains tabs inside of strings
+    // TODO: replace \t with two spaces
+    if(body.includes('\t')){
+      body = body.replace(/\t/g, '  ')
+    }
 
     try {
       debug('parse json')
--- a/test/json.js
+++ b/test/json.js
@@ -20,6 +20,14 @@ describe('bodyParser.json()', function () {
       .expect(200, '{"user":"tobi"}', done)
   })
 
+  it('should parse JSON that contains tabs', function (done) {
+    request(createServer())
+      .post('/')
+      .set('Content-Type', 'application/json')
+      .send('{"fullname":"Abderrahim	Oukhrib"}')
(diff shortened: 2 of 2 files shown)


wesleytodd:
Hi, can you provide more information on the behavior and error you are having? It appears you are trying to change a behavior from `JSON.parse` which is provided by the language runtime not this library. If that is the case (please tell me if I am wrong) we are not interested in this type of additional behavior.

jonchurch:
Wow TIL:
[ECMA 404 v2](https://ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) section 4
> Insignificant whitespace is allowed before or after any token. Whitespace is any sequence of one or more of
the following code points: character tabulation (U+0009), line feed (U+000A), carriage return (U+000D), and
space (U+0020). Whitespace is not allowed within any token, except that space is allowed in strings.

We are not able to accept this change because parsing invalid JSON is out of the scope of this package.

Both of these will throw because they are invalid JSON.

```js
JSON.parse('{"fullname":"John	Doe"}')
JSON.parse('{"fullname":"John \t Doe"}')
```
For completeness, this will not throw, because escaping the tab makes it valid:
```js
JSON.parse('{"fullname":"John \\t Doe"}')
```
This also will not throw, because control characters outside of values are not considered invalid
```js
JSON.parse('{"fullname": \t"John Doe"}')
```

This isn't just true for tabs, but any non space whitespace inside of a JSON value itself.

These also throw for the same reason:

``js
JSON.parse('{"fullname": "John \nDoe"}')
JSON.parse('{"fullname": "John \rDoe"}')
```

If you want to craft a valid JSON string from your JS object:

```js
JSON.stringify({"fullname":"John      Doe"})
// '{"fullname":"John\\tDoe"}'
```

PR #523, opened 2024-05-05
ci: fix errors in ci github action for node 8 and 9

inigomarquinez:
This PR fixes [nyc](https://github.com/istanbuljs/nyc) version to 14.1.1 when running tests in node 8 or node 9. `nyc 15.x `requires a `yargs` package version that requires node >=10.

I've also added the latest versions of node (20, 21 and 22) to the matrix.

Related to https://github.com/jshttp/http-errors/issues/108

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,80 +51,80 @@ jobs:
           npm-i: mocha@3.5.3 nyc@10.3.2 supertest@2.0.0
 
         - name: io.js 1.x
-          node-version: "1.8"
+          node-version: "1"
           npm-i: mocha@3.5.3 nyc@10.3.2 supertest@2.0.0
 
         - name: io.js 2.x
-          node-version: "2.5"
+          node-version: "2"
           npm-i: mocha@3.5.3 nyc@10.3.2 supertest@2.0.0
 
         - name: io.js 3.x
-          node-version: "3.3"
+          node-version: "3"
           npm-i: mocha@3.5.3 nyc@10.3.2 supertest@2.0.0
 
         - name: Node.js 4.x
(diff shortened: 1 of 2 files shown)


inigomarquinez:
It seems that the tests fail for node >= 21, but not related to this PR itself, but to node itself.

UlisesGascon (approved):
LGTM!

wesleytodd (approved):

wesleytodd:
Ah, I am surprised we have more tests in here failing because of the query thing, but that is unrelated here. @jonchurch did you end up getting this fixed in that other branch? If so, mind dropping a link here so we don't have to repeat work to resolve it.

wesleytodd:
I just tested locally and latest 22 does not fail. Do you want to update this to use `22.2.0`? I am actually unsure what benefit we get from not just taking the latest in the major line, could we maybe just drop the minor specifier as well?

inigomarquinez:
@wesleytodd , minor version removed from node 22. It still fails for node 21. Should I also removed minor version?

wesleytodd:
Yeah I think we should remove the minor from all of them. We really want to be testing the latest in each major line without having to update all of them all the time.

inigomarquinez:
@wesleytodd , I've removed the minor versions

@UlisesGascon , @carpasse , perhaps something to have in mind for all the other ci pipelines of the repositories we've been fixing/migrating 🤔

wesleytodd (approved):

ctcpip:
https://github.com/inigomarquinez/expressjs-body-parser/pull/1 addresses the failure in 21

jonchurch (approved):

PR #547, opened 2024-10-21
Remove devDependency `safe-buffer`

Phillip9587:
This PR removes the `safe-buffer` package from devDependencies as it's no longer needed. Since we no longer support Node.js versions older than 18.0.0, the native Buffer class now fully handles security concerns. Additionally, there are no instances of `new Buffer(...)` usage in the repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "methods": "1.1.2",
     "mocha": "10.2.0",
     "nyc": "15.1.0",
-    "safe-buffer": "5.2.1",
     "supertest": "6.3.3"
   },
   "files": [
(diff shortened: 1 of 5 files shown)


UlisesGascon (approved):
LGTM! I will add the @expressjs/security-wg add reviewer too.

UlisesGascon:
We probably want to reflect this change in the `HISTORY.md` file.

Phillip9587:
> We probably want to reflect this change in the `HISTORY.md` file.

Is it necessary for a devDependency?

UlisesGascon:
> Is it necessary for a devDependency?

Oh! you are right, forget my last comment :)

bjohansebas (approved):

PR #548, opened 2024-10-21
Remove unused devDependency `methods`

Phillip9587:
This PR removes the unused `methods` package from `devDependencies`. After reviewing the codebase, i think that `methods` is no longer required or used anywhere in the project.


UlisesGascon (approved):
Great catch, seems like we are no longer use methods. So we can remove it :+1:

PR #549, opened 2024-10-22
test: remove AsyncLocalStorage check

Phillip9587:
with the minimum supported node version set to v18 we can remove the AsyncLocalStorage support check as it was added in v13.10.0 and v12.17.0


UlisesGascon (approved):

bjohansebas (approved):
LGTM

PR #550, opened 2024-10-22
refactor: extract shared utility functions

Phillip9587:
This PR refactors the code by moving the `getCharset` and `typeChecker` functions, which were previously duplicated across different parts of the codebase, into a shared `lib/utils.js` file.

#### **Key Benefits:**
- **Code reuse**: Both functions are now centralized, reducing duplication.
- **Size optimization**: By eliminating repeated code, this change helps shave off some bytes.
- **Maintainability**: Future updates to these functions can be made in one place, simplifying maintenance.

No functionality has been altered—this is strictly a refactor for improved efficiency and modularity.

PR #544, opened 2024-10-16
rebase: Generic Body Parser implemented

ctcpip:
supersedes https://github.com/expressjs/body-parser/pull/524

PR #551, opened 2024-10-22
refactor: normalize common options for all parsers

Phillip9587:
Based on #550 

This PR introduces the `normalizeOptions` function, which standardizes and validates common parser options like `inflate`, `limit`, `type`, and `verify`. It ensures consistent defaults and removes duplication across the parsers.


wesleytodd:
How much of this would also overlap with @ctcpip's attempt to revive #544? Would we be better off talking that approach to simplification?

lib/utils.js
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * Normalizes the common options for all parsers.
+ *
+ * @param {object} options
+ * @param {string} defaultType
+ * @returns {object}
+ */
+function normalizeOptions (options, defaultType) {
```
jonchurch, line 61: We could throw if `defaultType` is not provided

Previously it was implicitly required in that all parsers defined a fallback, but the current approach will accept undefined here silently and make the parser match against any Content-Type header provided in a req. It's unlikely that is expected or desired behavior on the part of a maintainer creating a new parser.

I put it up top because we are verifying internal use here, not the normalized value after considering the user input.

```suggestion
function normalizeOptions (options, defaultType) {
  if (!defaultType) {
    // Parsers must define a default content type or function
    throw new TypeError('defaultType must be provided')
  }
```

jonchurch (changes requested):
it LGTM, but with the one nit I left so marking as request for change. The nit doesn't have to block, but I think it's a sane addition.

Phillip9587:
@jonchurch I've made the requested changes. I think we should merge this first as foundational work for #22, so that when we start on it, the necessary changes are already in place and the scope remains smaller. Let me know what you think!

lib/utils.js
```diff
@@ -0,0 +1,89 @@
+/*!
+ * body-parser
+ * Copyright(c) 2014-2015 Douglas Christopher Wilson
```
bjohansebas, line 3: It's a new file, so it would be better to omit this or add that it should be express contributors.

lib/utils.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * Normalizes the common options for all parsers.
+ *
+ * @param {object} options
+ * @param {string} defaultType
+ * @returns {object}
+ */
+function normalizeOptions (options, defaultType) {
```
bjohansebas, line 61: You probably want to add tests to check its behavior at the function level

lib/utils.js
Phillip9587: Should I remove the header completely?

bjohansebas: Probably this is the correct header only for this file, since we have to respect copyright rights. This happened on the website and the following copyright was added

```suggestion
 * Copyright(c) Express Contributors
```

bjohansebas (approved):
LGTM

jonchurch (changes requested):

lib/utils.js
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @param {object} options options to normalize
+ * @param {string} defaultType default content type
+ * @returns {object}
+ */
+function normalizeOptions (options, defaultType) {
+  if (!defaultType || typeof defaultType !== 'string') {
+    // Parsers must define a default content type
+    throw new TypeError('defaultType must be provided')
+  }
```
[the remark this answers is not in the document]
jonchurch: Woops, I missed something else here. 

User facing `type` option should ahve the signature of `string | string[] | function`

This is getting complicated for `defaultType` so Im going to back track here and resolve the comment.

We don't need to worry rn about validating the internal use, I just wanted to catch anyone who accidentally forgot to pass a `defaultType` when creating a new parser.

test/utils.js
```diff
@@ -0,0 +1,123 @@
...
+      }, /defaultType must be provided/)
+    })
+
+    it('should throw an error if defaultType is not a string', () => {
+      assert.throws(() => {
+        normalizeOptions({}, 123)
+      }, /defaultType must be provided/)
+    })
+  })
+})
```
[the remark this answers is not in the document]
jonchurch: backtracking here and marking as resolved, will suggest the functionality check in another comment

lib/utils.js
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Normalizes the common options for all parsers.
+ *
+ * @param {object} options options to normalize
+ * @param {string} defaultType default content type
+ * @returns {object}
+ */
+function normalizeOptions (options, defaultType) {
+  if (!defaultType || typeof defaultType !== 'string') {
```
jonchurch, line 62: Lets keep this simpler for now and just validate that defaultType is defined
```suggestion
  if (!defaultType) {
```

lib/utils.js
```diff
@@ -0,0 +1,89 @@
...
+  return function checkType (req) {
+    return Boolean(typeis(req, type))
+  }
+}
+
+/**
+ * Normalizes the common options for all parsers.
+ *
+ * @param {object} options options to normalize
+ * @param {string} defaultType default content type
```
jonchurch, line 58:
```suggestion
 * @param {string | string[] | function} defaultType default content type(s) or a function to determine it
```

test/utils.js
jonchurch, line 123: Added a test under describe type to test usage w/ array of strings
Removed the "should throw an error if defaultType not a string" test.

added two tests to  assert on the functionality of defaultType when provided as string, array of strings, or function

```suggestion
    describe('type', () => {
      it('should return the default type if type is not provided', () => {
        const result = normalizeOptions({}, 'application/json')
        assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/json', 'content-length': '1024' } }), true)
        assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/xml', 'content-length': '1024' } }), false)
      })

      it('should accept a string type', () => {
        const result = normalizeOptions({ type: 'application/xml' }, 'application/json')
        assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/xml', 'content-length': '1024' } }), true)
        assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/json', 'content-length': '1024' } }), false)
      })

      it('should accept an array of strings type', () => {
        const result = normalizeOptions({ type: ['application/xml', 'application/*+json'] }, 'application/json')
        assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/xml', 'content-length': '1024' } }), true)
        assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/ld+json', 'content-length': '1024' } }), true)
        assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/json', 'content-length': '1024' } }), false)
      })

      it('should accept a type checking function', () => {
        const result = normalizeOptions({ type: () => true }, 'application/json')
        assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/xml' } }), true)
        assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/json' } }), true)
      })
    })
  })

  describe('defaultType', () => {
    it('should throw an error if defaultType is not provided', () => {
      assert.throws(() => {
        normalizeOptions({})
      }, /defaultType must be provided/)
      assert.throws(() => {
        normalizeOptions({}, undefined)
      }, /defaultType must be provided/)
    })

    it('should convert string defaultType to a request content-type checking function', () => {
      const result = normalizeOptions({}, 'application/json')
      assert.strictEqual(typeof result.shouldParse, 'function')
      assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/json', 'content-length': '1024' } }), true)
      assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/xml', 'content-length': '100' } }), false)
    })

    it('should convert string defaultType to a request content-type checking function', () => {
      const result = normalizeOptions({}, 'application/json')
      assert.strictEqual(typeof result.shouldParse, 'function')
      assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/json', 'content-length': '1024' } }), true)
      assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/xml', 'content-length': '100' } }), false)
    })

    it('should use function defaultType directly as the request content-type checker', () => {
      const typeFunction = (req) => req.headers['content-type'].endsWith('+json')
      const result = normalizeOptions({}, typeFunction)
      assert.strictEqual(result.shouldParse, typeFunction)
      assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/ld+json' } }), true)
      assert.strictEqual(result.shouldParse({ headers: { 'content-type': 'application/xml' } }), false)
    })
  })
})
```

jonchurch (changes requested):

lib/utils.js
```diff
@@ -0,0 +1,89 @@
...
+    return undefined
+  }
+}
+
+/**
+ * Get the simple type checker.
+ *
+ * @param {string} type
+ * @return {function}
+ */
```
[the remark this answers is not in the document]
Phillip9587: I think that the type should be just `string | string[]` here. If a function is passed as `type` oder `defaultType` we just call that function and don't need the default type-checker

jonchurch (approved):

Phillip9587:
Hey everyone, I just wanted to follow up on this PR and see if there's anything I can do to help move it forward. Let me know if any changes or additional context are needed! Thanks!

Phillip9587:
Hey @jonchurch, thanks again for your review and approval a few weeks ago! I just wanted to check in and see if it’s all right to go ahead and merge this. Let me know if you have any additional concerns.

UlisesGascon:
Is this a semver minor, right?

bjohansebas:
I would say it's a patch, this change doesn't introduce any new functionality, it just refactors it.

bjohansebas (approved):

UlisesGascon (approved):

bjohansebas (approved):

Phillip9587:
@UlisesGascon Can we merge this now?

UlisesGascon:
I am checking with @wesleytodd... So far the plan is to include it in the next release (that will be shipped to in Express 5.1.0) :)

Phillip9587:
@UlisesGascon Thanks! My work on #22 depends on this, so I'd really appreciate it if it could be merged soon.

PR #555, opened 2024-10-31
Refactor firstchar function for conciseness and modern syntax

Ayoub-Mabrouk:
- Replaced var with const for improved variable management
- Used optional chaining for a more concise return statement
- Enhanced readability while preserving functionality

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -189,11 +189,8 @@ function createStrictSyntaxError (str, char) {
  */
 
 function firstchar (str) {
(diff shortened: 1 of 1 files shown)


bjohansebas (approved):

wesleytodd (changes requested):
I see approvals on this. We have never accepted style changes before, and I dont believe we should now. But this maybe a discussion we need to have at the TC level?

Phillip9587:
Closed by #744

PR #556, opened 2024-10-31
Refactor typeChecker function for conciseness and modern syntax

Ayoub-Mabrouk:
- Replaced traditional function declaration with an arrow function for brevity
- Utilized implicit return to streamline the code
- Enhanced readability while maintaining functionality


wesleytodd (changes requested):
This is a style change. We are not accepting style changes.

PR #562, opened 2024-11-19
perf: use the node require cache instead of custom caching

Phillip9587:
This PR replaces the custom caching logic with [Node.js's built-in require cache](https://nodejs.org/api/modules.html#caching). By leveraging the require cache, the codebase is simplified, performance is improved, and maintenance overhead is reduced. The custom logic has been removed, and the parser loading now relies on `require`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,13 +7,6 @@
 'use strict'
 
 /**
- * Cache of loaded parsers.
- * @private
- */
-
-var parsers = Object.create(null)
-
-/**
  * @typedef Parsers
  * @type {function}
  * @property {function} json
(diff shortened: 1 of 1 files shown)


wesleytodd (approved):
Generally a fan of this. I always wondered why this was done in the first place when lazy loading has worked like this for a LONG TIME.

wesleytodd:
Hey! No worries! I am just getting setup this week to get caught up (again, it is a never ending battle lol). I think this is good to merge, so will do so now. I am not going to prep a release just yet as I want to catch up on the ~500 notifications first to make sure we have the bigger picture plan in mind, but I will hopefully get releases ready for a bunch of this before the end of the month.

PR #565, opened 2024-12-06
ci: disable fail-fast in CI workflow

Phillip9587:
Added `fail-fast: false` to the job strategy matrix to ensure that the tests run on all node versions even if one of them fails.


bjohansebas (approved):

UlisesGascon (approved):

wesleytodd (approved):
THANK YOU! This is the most frustrating setting. We also had this in the test scripts in some cases so mocha would do this and I never understood it. If 5 tests fail in a related way I dont want to waste time thinking it is *one* test failing. Same as with node versions, why have to re-run it to see just another failure in a different node version after pusing?

PR #571, opened 2025-01-13
chore(deps): update `type-is` to v2.0.0

Phillip9587:
Upgrading `type-is` to v2 aligns with `express` v5, which depends on `type-is` v2. This alignment prevents duplication of the `type-is` package in `node_modules`.


wesleytodd (approved):

bjohansebas:
Would this cause any breaking changes? Looking at the history of type-is, it contains the following notes:

- Remove ambiguous extensions from IANA for application/*+xml types
- Update primary extension to .es for application/ecmascript

wesleytodd:
Specifically for body parser I dont think it will cause issues. I read the same changelog before approving and probably should have said that in a comment, sorry.

bjohansebas (approved):

Phillip9587:
Sorry I checked when i created the PR, but I should have documented it in the PR description.

PR #586, opened 2025-02-13
feat: zstandard support

### LINKED ISSUE #560: zstd support?
Now that we have brotli, can we also support zstd please?

Phillip9587:
This PR introduces support for the **Zstandard (Zstd) algorithm**, which was released as an experimental feature in [Node.js 23.8.0](https://nodejs.org/en/blog/release/v23.8.0#support-for-the-zstd-compression-algorithm).

## Key Changes
- Adds support for **Zstandard (Zstd)** as an optional compression algorithm.

## Notes
- This feature is experimental and subject to changes in future Node.js releases.
- Requires Node.js 23.8.0+ to function correctly.

--------------------------------------------------
closes #560

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ This module provides the following parsers:
   * [Text body parser](#bodyparsertextoptions)
   * [URL-encoded form body parser](#bodyparserurlencodedoptions)
 
+All parsers support automatic inflation of `gzip`, `br` (brotli), `deflate`
+and `zstd` encodings. `zstd` is available in Node.js `^22.15.0` and
+`>=23.8.0`.
+
 Other body parsers you might be interested in:
 
 - [body](https://www.npmjs.com/package/body#readme)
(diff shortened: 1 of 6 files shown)


bjohansebas (changes requested):
I would really prefer to wait until the API becomes stable. This is the same stance I’m going to take with *compression*. There may be no changes during the transition from experimental to stable, but it’s better to use only stable APIs

Great job anyway!

wesleytodd (changes requested):
I agree, we should wait until it is stable then just enable support without an experimental phase.

Phillip9587:
@UlisesGascon @jonchurch What are your opinions on this topic?

Phillip9587:
@UlisesGascon @wesleytodd All active LTS Node.js Versions now contain the Zstandard support. It is still experimental but already adopted by fastify in https://github.com/fastify/fastify-compress and by other ecosystem packages. I think we should also support it and document in which node.js versions zstd is available.

Phillip9587:
Hey @jonchurch @UlisesGascon, Can I please get some feedback on this?

bjohansebas (approved):
LGMT!

Phillip9587:
Hi @wesleytodd, @UlisesGascon, @jonchurch, could you please review this PR? I believe it would be great to get this moving. Thanks!

Phillip9587:
After asking for reviews since May 12 I'm just going to merge this now. We can still revert it if repo captains disagree

PR #587, opened 2025-02-15
refactor: optimize internal read function

Phillip9587:
This PR aims to improve the maintainability and readability of the internal read function. It also removes some weird parts of that code.

### DIFF
--- a/lib/read.js
+++ b/lib/read.js
@@ -64,61 +64,68 @@ function read (req, res, next, parse, debug, options) {
     return
   }
 
-  var encoding = null
+  var charset = null
   if (options?.skipCharset !== true) {
-    encoding = getCharset(req) || options.defaultCharset
+    charset = getCharset(req) || options.defaultCharset
 
     // validate charset
-    if (!!options?.isValidCharset && !options.isValidCharset(encoding)) {
+    if (!!options?.isValidCharset && !options.isValidCharset(charset)) {
       debug('invalid charset')
-      next(createError(415, 'unsupported charset "' + encoding.toUpperCase() + '"', {
-        charset: encoding,
+      next(createError(415, 'unsupported charset "' + charset.toUpperCase() + '"', {
+        charset,
         type: 'charset.unsupported'
       }))
       return
     }
   }
 
-  var length
-  var opts = options
-  var stream
+  // get the content stream
+  const contentEncoding = (req.headers['content-encoding'] || 'identity').toLowerCase()
+  debug('content-encoding "%s"', contentEncoding)
 
-  // read options
-  var verify = opts.verify
-
-  try {
-    // get the content stream
-    stream = contentstream(req, debug, opts.inflate)
-    length = stream.length
-    stream.length = undefined
-  } catch (err) {
-    return next(err)
+  if (options.inflate === false && contentEncoding !== 'identity') {
+    return next(createError(415, 'content encoding unsupported', {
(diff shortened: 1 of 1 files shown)


lib/read.js
```diff
@@ -36,49 +36,55 @@ module.exports = read
...
-  var verify = opts.verify
-
-  try {
-    // get the content stream
-    stream = contentstream(req, debug, opts.inflate)
-    length = stream.length
-    stream.length = undefined
-  } catch (err) {
-    return next(err)
+  const charset = options.charset
```
Phillip9587, line 40: I renamed the option to `charset` because it holds the `content-type` header charset or the default charset

lib/read.js
```diff
@@ -133,39 +139,6 @@ function read (req, res, next, parse, debug, options) {
...
-
-  if (encoding === 'identity') {
-    req.length = length
-    return req
-  }
-
-  var stream = createDecompressionStream(encoding, debug)
-  req.pipe(stream)
-  return stream
-}
```
Phillip9587, line 167: This can be seen as follow-up to #564. I included the `contentstream` function in the main `read` function for multiple reasons:
- to prevent the property `length` from being set and reset on the `req` object. 
- This also improves code by only setting the `raw-body` `length` option and thus also only calling `req.headers['content-length']` when the requests needs no inflation.

lib/read.js
```diff
@@ -36,49 +36,55 @@ module.exports = read
...
-  // set raw-body options
-  opts.length = length
-  opts.encoding = verify
-    ? null
-    : encoding
+  let stream
+  if (contentEncoding === 'identity') {
+    // set raw-body expected length
+    stream = req
+    options.length = req.headers['content-length']
```
Phillip9587, line 57: Improves the code by only accessing `req.headers['content-length']` and setting the `length` option for `raw-body` when the stream needs no decompression

lib/types/urlencoded.js
```diff
@@ -111,13 +109,10 @@ function urlencoded (options) {
...
-      encoding: charset,
-      inflate: inflate,
-      limit: limit,
-      verify: verify,
-      charsetSentinel: charsetSentinel,
-      interpretNumericEntities: interpretNumericEntities
+      charset,
+      inflate,
+      limit,
+      verify
```
Phillip9587, line 115: We only need the 4 options here

Phillip9587 (commented):
I’ve added comments to explain my changes. The diffs don’t fully capture the modifications clearly, so I’d recommend comparing the file from master with the version in this branch for better context.

Let me know if you have any questions or need further clarifications. Looking forward to your feedback! 😊

lib/read.js
```diff
@@ -36,49 +36,55 @@ module.exports = read
...
 
 function read (req, res, next, parse, debug, options) {
-  var length
-  var opts = options
-  var stream
-
   // read options
-  var encoding = opts.encoding !== null
-    ? opts.encoding
-    : null
```
Phillip9587, line 46: I think we can remove that 😅

lib/types/urlencoded.js
Phillip9587: Updated in another PR

Phillip9587:
Hey @UlisesGascon @jonchurch, just circling back on this PR - it's been a little while, so I wanted to check in and see if you had a chance to take a look. Let me know if there's anything I should address or update to help move it forward. Thanks!

PR #589, opened 2025-02-18
fix: remove skip of test

bjohansebas:
We have a test that was being skipped, and I don't think there was any reason why it should be skipped.


Phillip9587 (approved):

jonchurch (approved):

jonchurch:
fast track merging as its a small test fix, dont want this to break anything ✅

PR #591, opened 2025-02-23
Refactor parameterCount to optimize performance

wojtekmaj:
Now that body-parser targets Node.js 18, we can finally use String.split to count all the `&`s in `parameterCount` body. This makes the function roughly 2x faster in Chrome/Node.js engine (and about 2.5x faster in Safari/Bun engine), according to this jsperf: https://jsperf.app/niqepi/2

Fun fact: if you run this jsperf in Firefox, the old implementation will get speeds unmatched by any other engine, while the _new_ one will be roughly 7x slower. Thankfully, that's only the case in Firefox.

a

### DIFF
--- a/lib/types/urlencoded.js
+++ b/lib/types/urlencoded.js
@@ -210,7 +210,7 @@ function getCharset (req) {
 function parameterCount (body, limit) {
   var len = body.split('&').length
 
-  return len > limit ? undefined : len - 1;
+  return len > limit ? undefined : len - 1
 }
 
 /**
--- a/lib/types/urlencoded.js
+++ b/lib/types/urlencoded.js
@@ -208,19 +208,9 @@ function getCharset (req) {
  */
 
 function parameterCount (body, limit) {
-  var count = 0
-  var index = 0
+  var len = body.split('&').length;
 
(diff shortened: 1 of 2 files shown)


UlisesGascon (approved):
Oh yeah!

lib/types/urlencoded.js
```diff
@@ -208,19 +208,9 @@ function getCharset (req) {
  */
 
 function parameterCount (body, limit) {
-  var count = 0
-  var index = 0
+  var len = body.split('&').length;
```
UlisesGascon, line 211:
```suggestion
  var len = body.split('&').length
```

lib/types/urlencoded.js
```diff
@@ -208,19 +208,9 @@ function getCharset (req) {
...
-    count++
-    index++
-
-    if (count === limit) {
-      return undefined
-    }
-  }
-
-  return count
+  return len > limit ? undefined : len - 1;
```
UlisesGascon, line 213:
```suggestion
  return len > limit ? undefined : len - 1
```

UlisesGascon (commented):
Let's make the linter happy again :)

PR #595, opened 2025-03-26
ci: remove unnecessary checkout step from coverage job

Phillip9587:


Phillip9587:
It seems that the coveralls action depends on the checkout step.

PR #598, opened 2025-03-27
replace `on-finished` with native `stream.finished` for request handling

bjohansebas:
I'm investigating how to stop using `on-finished` here and handle it directly with the properties that Node.js provides. I don't know if it's possible or if there are any bugs.  

It's likely that changes will need to be made in raw-body, as it handles a large part of this module's logic.

### DIFF
--- a/lib/read.js
+++ b/lib/read.js
@@ -14,8 +14,8 @@
 var createError = require('http-errors')
 var getBody = require('raw-body')
 var iconv = require('iconv-lite')
-var onFinished = require('on-finished')
 var zlib = require('node:zlib')
+const { finished } = require('node:stream')
 
 /**
  * Module exports.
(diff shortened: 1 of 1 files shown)


lib/read.js
```diff
@@ -201,10 +201,10 @@ function createDecompressionStream (encoding, debug) {
  */
 
 function dump (req, callback) {
-  if (onFinished.isFinished(req)) {
+  if (req.writableEnded || req.writabledFinished) {
     callback(null)
   } else {
-    onFinished(req, callback)
```
bjohansebas, line 207: I've opened an issue in Node to fix the tests that are failing due to this change.

https://github.com/nodejs/node/issues/57844

bjohansebas:
on-finished stays, since https://github.com/nodejs/node/issues/57844 is a bug it can't be integrated, and isFinished checks too many things to keep it here, so let's stick with that package.

PR #600, opened 2025-03-27
refactor: move common request validation to read function

Phillip9587:
This PR moves the common request validation logic, previously duplicated across all parsers, into the `read` function. This improves maintainability, reduces package size, and brings us closer to our goal of a generic body parser (see #22).

It also adds support for the `defaultCharset` option in `normalizeOptions()`. This option was already supported in the `qs` and `text` parsers, and I’ve now added support in the `json` parser as well. While we’ll likely want to validate these options more thoroughly in a future major release, the current approach should suffice for now.

There’s not much left before we can move the full middleware creation logic into its own function, but I wanted to keep this PR focused and manageable. Small steps toward our goals.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+unreleased
+=========================
+
+* refactor: move common request validation to read function
+
 2.2.0 / 2025-03-27
 =========================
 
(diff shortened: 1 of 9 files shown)


bjohansebas (approved):

UlisesGascon (approved):
Thanks for this work @Phillip9587 :heart:

PR #601, opened 2025-03-31
chore: add .editorconfig

Phillip9587:
This pull request adds a `.editorconfig` file. The config is copied from [nodejs/node](https://github.com/nodejs/node/blob/5812a61a68d50c65127beb68dd4dfb0242e3c5c9/.editorconfig).

Docs: https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+quote_type = single


bjohansebas (changes requested):
I really don't want more configuration files, I don't see any benefit in it.  

I know it's also in the Express repository, but I honestly don't see any benefit and would even remove it from there as well.

Phillip9587:
Since we don’t currently have a linter or formatter in place across the repos, and especially in this one, which I’ve been working with extensively, I’d like to suggest adding an .editorconfig file.

This would help standardize code formatting across different editors and ensure consistent practices for things like indentation, line endings, and spacing. It can also help reduce unnecessary merge conflicts and make collaboration smoother overall.

With an .editorconfig file, contributors automatically follow a consistent style, regardless of the editor or IDE they use. It’s a small addition that can make a big difference.

bjohansebas:
We have a linter, which is Standard, and it also applies certain formatting. The previous maintainer implemented it in most, if not all, repositories.

We want to standardize a new linter because Standard no longer works well with the latest version of ESLint, and it doesn't seem like the Standard project will continue. That's why we're discussing the linter, but in reality, we already have a linter that needs to be updated because Standard won't be updated anymore.

Phillip9587:
@bjohansebas Maybe we can discuss in https://github.com/expressjs/discussions/issues/327

PR #602, opened 2025-04-04
refactor: move req.body = undefined to just before read attempt

smonn:
Reasoning: the various parsers should not modify the request object until it is clear it should be applied. Setting a property, even to undefined, may affect assumptions elsewhere. For example, `'req' in body` would be truthy even if the rest of the middleware was not applied.

Arguably, it can be removed entirely since all tests are still passing without it. But keeping it around in case there are other historical reasons behind it.

In either case, the current version is not compatible with, for example, tRPC v11 when using FormData. So without this change, the workaround is to add a condition to avoid applying the middleware at all. Example:

```js
const json = express.json();
app.use((req, res, next) => {
  if (req.url.includes('trpc')) return next();
  return json(req, res, next);
});
```

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -90,10 +90,6 @@ function json (options) {
       return
     }
 
-    if (!('body' in req)) {
-      req.body = undefined
-    }
-
     // skip requests without bodies
     if (!typeis.hasBody(req)) {
       debug('skip empty body')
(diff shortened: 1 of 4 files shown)


Phillip9587 (approved):
Thanks for your contribution! Personally, I'm in favor of removing it completely, but let's wait to hear what the others think. I'm not entirely sure if this would be considered a breaking change, since some users might be relying on this behavior.

bjohansebas (approved):
LGTM

bjohansebas:
> I'm in favor of removing it completely,

I'm neutral on this, i'd need to do a bit more research to see if this validation is really necessary.

Issue #347, opened 2019-02-08
Implement a __proto__ check option


martinheidegger:
Eran Hammer posted an article on __proto__ poisoning and his solution in joi/hapi: https://hueniverse.com/a-tale-of-prototype-poisoning-2610fa170061

@rgrove posted a simple implementation of a fix for this: https://gist.github.com/rgrove/3ea9421b3912235e978f55e291f19d5d

However the fix requires a custom reviver that might slow down the default/valid parsing case, Eran prevented this by using an initial check for `__proto__`. It might be good to add this as a default to be checked for in body-parser in general that can be switched off... if someone wants to do so.

dougwilson:
We are working on this as a security issue that was reported.

dougwilson:
In general unless there is a consensus that JSON.parse is fundamentally unsafe and should not be used, there isn't actually a security issue at had in this module. The security issue highlighted there is the nature of the language itself, and one could argue that Object.assign is the vulnerable point, as that is the point of failure.

P.S. I didn't mean to close the issue

dougwilson:
One method to protect your app if you have vulnerable code (and are not willing to fix it), would be to add the following middleware after body parser:

```js
app.use((req, res, next) => {
  if (req.body && typeof(req.body === 'object') Bourne.scan(req.body, { your options })
  next()
})
```

This uses the https://github.com/hapijs/bourne instead of a reviver. Not sure if that is faster or not performance wise, off hand.

martinheidegger:
> In general unless there is a consensus that JSON.parse is fundamentally unsafe and should not be used, there isn't actually a security issue at had in this module. The security issue highlighted there is the nature of the language itself, and one could argue that Object.assign is the vulnerable point, as that is the point of failure.

I was thinking the same thing. Object.assign is definitely inconsistent here. Lets prevent developers from using Object.assign 😉  - ! Sadly not a practical solution.

The argument for fixing this in `body-parser` to me is regarding unsanitized input. It can be perfectly fine that JSON data contains a __proto__ object if the owner and intents are okay. If an app feeds Json.parse a unsanitized string, however it should check for __proto__. As `body-parser` is basically a glorified parser for unsanitized content, it seemed like the right repository to open this issue.

dougwilson:
Sure, but there are many other issues. For example, even `constructor` should arguably be sanitized out, as modules similar to `Object.assign` would pollute the prototype with that property name: https://hackerone.com/reports/430291

[...]

PR #603, opened 2025-04-05
feat: add secure JSON parsing with prototype poisoning handling

bjohansebas:
Users of this middleware are given the option to control prototype poisoning
closes: #347

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -18,6 +18,7 @@ var isFinished = require('on-finished').isFinished
 var read = require('../read')
 var typeis = require('type-is')
 var { getCharset, normalizeOptions } = require('../utils')
+const secureJson = require('secure-json-parse')
 
 /**
  * Module exports.
@@ -55,6 +56,7 @@ function json (options) {
 
   var reviver = options?.reviver
   var strict = options?.strict !== false
+  const protoPoisoning = options?.onProtoPoisoning || 'ignore'
 
   function parse (body) {
     if (body.length === 0) {
@@ -74,7 +76,7 @@ function json (options) {
 
     try {
       debug('parse json')
-      return JSON.parse(body, reviver)
+      return secureJson.parse(body, reviver, { protoAction: protoPoisoning })
     } catch (e) {
       throw normalizeJsonSyntaxError(e, {
         message: e.message,
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "on-finished": "^2.4.1",
     "qs": "^6.14.0",
     "raw-body": "^3.0.0",
+    "secure-json-parse": "^4.0.0",
     "type-is": "^2.0.0"
   },
   "devDependencies": {
--- a/test/json.js
+++ b/test/json.js
@@ -721,6 +721,32 @@ describe('bodyParser.json()', function () {
       test.expect(413, done)
     })
   })
+
+  describe("prototype poisoning", function () {
+    it('should parse __proto__ when protoAction is set to ignore', function (done) {
+      request(createServer({ onProtoPoisoning: 'ignore' }))
+        .post('/')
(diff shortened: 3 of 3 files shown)


bjohansebas:
[secure-json-parse](https://github.com/fastify/secure-json-parse) also has protection against constructor poisoning. Should I include that as well? Are there any use cases for it?

lib/types/json.js
```diff
@@ -55,6 +56,7 @@ function json (options) {
 
   var reviver = options?.reviver
   var strict = options?.strict !== false
+  const protoPoisoning = options?.onProtoPoisoning || 'ignore'
```
bjohansebas, line 59: In a future major version, we can make this throw an error or ignore it, but in the meantime, to maintain compatibility, this option is handled.

package.json
```diff
@@ -17,6 +17,7 @@
     "on-finished": "^2.4.1",
     "qs": "^6.14.0",
     "raw-body": "^3.0.0",
+    "secure-json-parse": "^3.0.2",
```
bjohansebas, line 20: There's a version 4, but they theoretically support from version 20 onwards, so to avoid any potential issues, we’re sticking with version 3, which supports Node 18+

lib/types/json.js
```diff
@@ -74,7 +76,7 @@ function json (options) {
 
     try {
       debug('parse json')
-      return JSON.parse(body, reviver)
+      return secureJson.parse(body, reviver, { protoAction: protoPoisoning })
```
bjohansebas, line 79: Another option would be to always protect against this issue by passing 'error' or 'remove', but I'm not sure. Other packages chose to make it a configurable option, and I think it would make some sense to do the same.

Phillip9587, line 79:
```suggestion
      return secureJson.parse(body, reviver, { protoAction: protoPoisoning, constructorAction: 'ignore' })
```

You have to set `constructorAction` to `'ignore'` to not introduce a breaking change.

Method args: https://github.com/fastify/secure-json-parse?tab=readme-ov-file#sjsonparsetext-reviver-options

Phillip9587 (changes requested):

bjohansebas:
Yeah, although I don't think it hurts to add an option for constructor poisoning, so I'm going to add tests for the constructor prototype and add the option

lib/types/json.js
bjohansebas: fixed

jonchurch:
Im not crazy about the current form of the PR. I'd prefer we fight against bloating the options further, especially in cases where there is no strong reason for someone to deviate from the default.

Philosophically I don't like these kinds of mitigations for language level issues, they should be fixed in the language. That ship has long since sailed, as `Object.assign` is likely never to change. At least the spread operator was fixed to no longer set prototype.

I'd consider dropping `__proto__, constructor.prototype` (and possibly even `constructor` outright) from user inputs silently in a major, and not providing an option for people to toggle it off. If someone needs to deviate that far, they can extend the Generic parser to do it.

bjohansebas:
> I'd consider dropping __proto__, constructor.prototype (and possibly even constructor outright) from user inputs silently in a major, and not providing an option for people to toggle it off. If someone needs to deviate that far, they can extend the Generic parser to do it.

I agree with that approach, adding the option was more about being permissive, but I’d really prefer to just remove it.

jonchurch:
Btw I was wrong about:

> the extended urlencoded parser (using qs) already drops proto, and there is no way to turn that off. The non extended urlencode uses querystring and would still pass through proto.

1. non-extended went to `qs` in 2.0.0
2. qs is being configured with `allowPrototypes: true`, but there is no option to opt out of that. 

So for the consistency argument here when landing a proto pollution mitigation to `bodyParser.json` that was always on, we'd toggle the qs option to false to be consistent

PR #604, opened 2025-04-05
refactor: do not overwrite the error stack in json parser

bjohansebas:
to simplify the JSON parser a bit

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -77,8 +77,7 @@ function json (options) {
       return JSON.parse(body, reviver)
     } catch (e) {
       throw normalizeJsonSyntaxError(e, {
-        message: e.message,
-        stack: e.stack
+        message: e.message
       })
     }
   }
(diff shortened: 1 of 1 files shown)


jonchurch:
I didn't realize this stack manipulation was here tbh. Likely no one is matching on stacktraces from body-parser, but with things that have potentially been part of someone's monitoring stack for 8 years, Im hesitant to change it outside a major.

I'd rather we keep this how it is until we decide to change error reporting in this package in a more meaningful way ala using [`error.cause`](https://nodejs.org/api/errors.html#errorcause) as a better way to enrich errors while keeping the stack traceable

Phillip9587:
@bjohansebas Can you please open an issue so we don't loose it for the next major version?

bjohansebas:
i'd say this is a clean change, since it only modifies the error message, which is already being handled in the error.message part, and it wouldn't introduce any breaking changes.

and it does make sense to migrate it to error.cause later, i’ll open an issue for that.

PR #606, opened 2025-04-08
Generic Parser

jonchurch:
An attempt at solving for #22 and sketching out what the generic interaface would look like.

This is a draft so we can see the diff cleanly, but it's my intent for Generic ot be hashed out on the `generic` branch.

This is an approach which encapsulates Generic as the only middleware to have a `parser` option. 

Users of generic would be required to bring a `type` and a sync `parse` function

### DIFF
--- a/lib/generic.js
+++ b/lib/generic.js
@@ -1,0 +1,149 @@
+/*!
+ * body-parser
+ * Copyright(c) 2014 Jonathan Ong
+ * Copyright(c) 2014-2015 Douglas Christopher Wilson
+ * MIT Licensed
+ */
+
+'use strict'
+
+/**
+ * Module dependencies.
+ * @private
+ */
+
+const contentType = require('content-type')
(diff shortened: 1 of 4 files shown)


lib/generic.js
```diff
@@ -0,0 +1,153 @@
...
+ * @param {Boolean} [options.inflate] Enable handling compressed bodies
+ * @param {Function} [options.verify] Verify body content
+ * @param {String} [options.defaultCharset] Default charset when not specified
+ * @param {String} [options.charset] Expected charset (will respond with 415 if not matched)
+ * @return {Function} middleware
+ * @public
+ */
+
+function generic (options) {
+  // === STEP 0: VALIDATE OPTIONS ===
```
jonchurch, line 51: Breaking this function down into steps was very helpful for me when wrapping my head around what was a universal concern vs parser specfic. leaving them here hoping they help others

PR #631, opened 2025-08-19
deps: bump iconv-lite

bjohansebas:
See https://github.com/ashtuchkin/iconv-lite/pull/334. Since this is like a new major version of the package, Dependabot won’t update it, and also to see how it behaves.

The only change that affects body-parser is https://github.com/ashtuchkin/iconv-lite/pull/328.


Phillip9587 (approved):

Phillip9587 (changes requested):
Please add a HISTORY entry

PR #651, opened 2025-10-28
perf: eliminate conditional check in json strict mode hot path

Phillip9587:
This change refactors the JSON parser to eliminate the per-request `if (strict)` conditional check from the parsing hot path.

Instead of checking strict mode during parsing, we now create specialized parser functions at middleware initialization based on the strict option.

I'm aware that this introduces some code duplication between the strict and non-strict parser implementations, but I think that is a necessary trade-off.

### DIFF
--- a/lib/types/json.js
+++ b/lib/types/json.js
@@ -48,23 +48,61 @@ var JSON_SYNTAX_REGEXP = /#+/g
 function json (options) {
   const normalizedOptions = normalizeOptions(options, 'application/json')
 
-  var reviver = options?.reviver
-  var strict = options?.strict !== false
+  const parse = createJsonParser(options)
 
-  function parse (body) {
-    if (body.length === 0) {
-      // special-case empty json body, as it's a common client-side mistake
-      // TODO: maybe make this configurable or part of "strict" option
-      return {}
-    }
+  const readOptions = {
+    ...normalizedOptions,
+    // assert charset per RFC 7159 sec 8.1
+    isValidCharset: (charset) => charset.slice(0, 4) === 'utf-'
+  }
 
-    if (strict) {
-      var first = firstchar(body)
+  return function jsonParser (req, res, next) {
+    read(req, res, next, parse, debug, readOptions)
+  }
+}
 
+/**
+ * Create a JSON parse function
+ *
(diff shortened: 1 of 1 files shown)


Phillip9587:
Hey @UlisesGascon @jonchurch 👋
Just revisited this PR and wanted to give it another gentle bump - it’s been a little while, and I’d love to get your feedback whenever you get a chance. Appreciate your time! 😊

Phillip9587:
Hey @UlisesGascon @jonchurch. A review would be appreciated

Phillip9587:
@UlisesGascon @jonchurch I plan to merge this next week. Maybe you have some time to check this out.

jonchurch (approved):
willing to give it a shot, if the duplication is still bothering you later we can land this where we extract out the biggest chunk that doesnt change per parser (make a `parseBody` function):

```js
function parseBody (body, reviver) {
  try {
    debug('parse json')
    return JSON.parse(body, reviver)
  } catch (e) {
    throw normalizeJsonSyntaxError(e, {
      message: e.message,
      stack: e.stack
    })
  }
}

/**
 * Create a JSON parse function
 *
 * @param {object} [options]
 * @return {function}
 * @private
 */
function createJsonParser (options) {
  const reviver = options?.reviver
  const strict = options?.strict !== false

  if (strict) {
    return function parseStrict (body) {
      if (body.length === 0) {
        // special-case empty json body, as it's a common client-side mistake
        // TODO: maybe make this configurable or part of "strict" option
        return {}
      }

      const first = firstchar(body)
      if (first !== '{' && first !== '[') {
        debug('strict violation')
        throw createStrictSyntaxError(body, first)
      }

      return parseBody(body, reviver)
    }
  }

  return function parse (body) {
    if (body.length === 0) {
      // special-case empty json body, as it's a common client-side mistake
      // TODO: maybe make this configurable or part of "strict" option
      return {}
    }

    return parseBody(body, reviver)
  }
}
```

Phillip9587:
> willing to give it a shot, if the duplication is still bothering you later we can land this where we extract out the biggest chunk that doesnt change per parser (make a `parseBody` function):

Let's maybe add this in a follow up PR.

PR #652, opened 2025-10-29
perf(urlencoded): optimize parameter counting for better memory efficiency

Ayoub-Mabrouk:
The previous implementation used `body.split('&')` which always processed the entire request body and allocated a full array, regardless of the parameter limit.

The new implementation:
- Counts '&' characters iteratively without array allocation
- Exits immediately when the limit is reached
- Reduces time complexity from O(n) worst-case always to O(min(n, limit))
- Added test case to verify empty body handling with `parameterLimit` option.

This particularly improves resilience against malicious requests with thousands of parameters attempting to exhaust server resources.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,7 +1,7 @@
-unreleased
+2.2.1 / 2025-11-24
 =========================
 
-* refactor: move common request validation to read function
+* Security fix for [GHSA-wqch-xfxh-vrr4](https://github.com/expressjs/body-parser/security/advisories/GHSA-wqch-xfxh-vrr4)
 * deps: 
   * type-is@^2.0.1
   * iconv-lite@^0.7.0
--- a/README.md
+++ b/README.md
@@ -483,14 +483,12 @@ app.use(bodyParser.text({ type: 'text/html' }))
 
 [MIT](LICENSE)
 
-[ci-image]: https://badgen.net/github/checks/expressjs/body-parser/master?label=ci
+[ci-image]: https://img.shields.io/github/actions/workflow/status/expressjs/body-parser/ci.yml?branch=master&label=ci
 [ci-url]: https://github.com/expressjs/body-parser/actions/workflows/ci.yml
-[coveralls-image]: https://badgen.net/coveralls/c/github/expressjs/body-parser/master
+[coveralls-image]: https://img.shields.io/coverallsCoverage/github/expressjs/body-parser?branch=master
 [coveralls-url]: https://coveralls.io/r/expressjs/body-parser?branch=master
-[node-version-image]: https://badgen.net/npm/node/body-parser
-[node-version-url]: https://nodejs.org/en/download
-[npm-downloads-image]: https://badgen.net/npm/dm/body-parser
+[npm-downloads-image]: https://img.shields.io/npm/dm/body-parser
 [npm-url]: https://npmjs.org/package/body-parser
-[npm-version-image]: https://badgen.net/npm/v/body-parser
+[npm-version-image]: https://img.shields.io/npm/v/body-parser
 [ossf-scorecard-badge]: https://api.scorecard.dev/projects/github.com/expressjs/body-parser/badge
 [ossf-scorecard-visualizer]: https://ossf.github.io/scorecard-visualizer/#/projects/github.com/expressjs/body-parser
--- a/lib/types/urlencoded.js
+++ b/lib/types/urlencoded.js
@@ -127,16 +127,23 @@ function createQueryParser (options) {
  *
  * @param {string} body
  * @param {number} limit
+ * @return {number|undefined} Returns undefined if limit exceeded
  * @api private
  */
-
 function parameterCount (body, limit) {
-  let count = 0
-  for (const char of body) {
-    if (char === '&') {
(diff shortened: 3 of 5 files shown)


bjohansebas:
Thank you for the contribution, this has already been resolved in our recent security patch.

https://github.com/expressjs/body-parser/security/advisories/GHSA-wqch-xfxh-vrr4

blakeembrey:
Should it just defer to `qs` and handle the error by enabling this option? https://github.com/ljharb/qs/blob/e8b32388dd8d095def77f5f21d3fdb7ca0c49cbc/lib/parse.js#L72-L74

ljharb:
I could implement a change in qs behind an option, that express sets - i'm unclear on what the change would be though.

Ayoub-Mabrouk:
> @blakeembrey Should it just defer to qs and handle the error by enabling this option? https://github.com/ljharb/qs/blob/e8b32388dd8d095def77f5f21d3fdb7ca0c49cbc/lib/parse.js#L72-L74

Hi @blakeembrey, I implemented your suggestion. Just to give some context, I had opened PR #652 back in October to fix the DoS issue with parameterCount that I mentioned in my earlier comment. @Phillip9587 committed the security patch (b204886) before my PR could be reviewed, which was good to get the fix out quickly. However, the implementation that got merged had a bug in the counting logic (the do-while loop issue I mentioned earlier, it counts iterations instead of actual `&` characters, returning incorrect values). So this PR is fixing that bug in the merged code and adding the missing tests.

Here's what I found:

I refactored the code to defer parsing to qs as you suggested, but kept the early exit since it's important for DoS protection. The new implementation uses `countAndValidateParameters()` that throws immediately when the limit is exceeded (before qs gets called), and then qs handles all the parsing work. Since we already validated, qs uses its default `throwOnLimitExceeded: false` which is what we want.

After implementing this, I realized the existing code was already doing most of what you suggested: it had early exit (qs never got called if limit exceeded), it was already deferring parsing to qs (since the default is false, qs doesn't check again), and the behavior was already optimal.

I tried the pure deferral approach (removing the manual counting, using `throwOnLimitExceeded: true`), but that would be slower because qs would do a full string split even when the limit's exceeded, which loses the early exit benefit that's important for DoS protection.

So the new code maintains the same optimal behavior as before, just with cleaner structure. It defers parsing to qs as you suggested, while keeping the early exit for performance.

---

> @bjohansebas Probably it would be good to delegate it to qs, but there would be a breaking change since it would no longer throw that error, which would most likely be for a major version. Additionally, it would be good to run benchmarks with and without using parameterCount and see how much it improves or worsens, because I feel we're repeating work that qs already does for us.

@bjohansebas I did some quick timing tests and found that keeping the early exit is important. When the limit is exceeded, the pure deferral approach (using `throwOnLimitExceeded: true`) would be slower because qs does a full string split before throwing, while the early exit approach avoids that entirely. For DoS protection, that early exit matters.

Regarding the "repeating work" concern, the existing code already had this optimization, and the new code keeps it while still deferring parsing to qs. So we're not really duplicating work: we do a quick pre-check for early exit, then qs handles all the parsing. If you'd like, I can run more thorough benchmarks to quantify the difference.

---

> @ljharb I could implement a change in qs behind an option, that express sets - i'm unclear on what the change would be though.

@ljharb Thank you for offering! After implementing this, I don't think we need any changes to qs. The existing `throwOnLimitExceeded` option works perfectly for this, we just use the default (`false`) since we already validate the limit before calling qs. So no changes needed on your end.

What do you all think? Is keeping the early exit while deferring parsing what you had in mind, or would you prefer going full deferral despite the performance trade-off?

Ayoub-Mabrouk:
@blakeembrey The function still returns incorrect values though - `parameterCount("", 1000)` returns `1` instead of `0`. The security patch's `do-while` loop introduced this regression by counting loop iterations instead of actual `&` characters.

I know it doesn't affect current behavior, but since it's a regression from the security patch, I thought it'd be worth fixing. The fix is straightforward and the PR adds test coverage for edge cases that weren't tested before. It keeps the function correct even if the current code doesn't depend on that correctness.

Would you be open to merging this? It's a small fix that corrects the regression and adds the missing test coverage.

Phillip9587:
> > This affects the arrayLimit calculation.
> 
> Actually it doesn't, it's `Math.max(paramCount, 100)` so it'll always be 100.
> 
> The only plausible place this affects anything is with `limit === 0` and allowing an empty string, but the function doesn't even allow setting 0 as a check earlier throws on `parameterLimit < 1`. As far as I can tell this fix does nothing.

I agree with Blake. This fix is not needed

Ayoub-Mabrouk:
> The security patch's `do-while` loop introduced this regression by counting loop iterations instead of actual `&` characters.

@Phillip9587  The security patch's do-while loop introduced this regression by counting loop iterations instead of actual & characters.  since it's a regression from the security patch I thought it'd be worth fixing

>This is the case that needs some fixup i think. But it is not critical as arrayLimit is set to Math.max(100, parameterCount).
I already talked to @UlisesGascon and he suggested opening a PR which adds tests and documentation for the expected behaviour of this function.

i guess you already agreed in the past to proceed to do the fix

Issue #687, opened 2026-01-03
`qs@6.14.1` update changes default `body-parser` behaviour


efekrskl:
Passing `arrayLimit` as 0 to `qs@6.14.1` parses arrays as objects with indexes instead of arrays (`lib/types/urlencoded.js:99`). Since `extended` is false by default, our default behaviour has changed.

There's also a failing test

```sh
  1) bodyParser.urlencoded()
       with extended option
         when false
           should parse multiple key instances:
     Error: expected '{"user":["Tobi","Loki"]}' response body, got '{"user":{"0":"Tobi","1":"Loki"}}'

```

## Environment information

**Version**: 2.2.1
<!-- 
To find the installed version of a package, you can check the `package.json` file in the root directory of your project. The version will be listed under `dependencies` or `devDependencies`, like this:

```json
"dependencies": {
  "express": "4.18.2"
}
-->

**Platform**:
<!-- 
UNIX: output of uname -a
Windows: output of "$([Environment]::OSVersion.VersionString) $(('x86', 'x64')[[Environment]::Is64BitOperatingSystem])" in PowerShell console 
-->

**Node.js version**:
<!-- 
Output of node -v.
-->

**Any other relevant information**:

## What steps will reproduce the bug?
`npm run test`

krzysdz:
According to https://github.com/ljharb/qs/issues/537#issuecomment-3705592022 this should have never worked in the first place. `body-parser` 2.0.0 changed the `"simple"` query parser to use `qs` instead of `querystring` (PR https://github.com/expressjs/body-parser/pull/387), but the behaviour should be the same (see discussion in https://github.com/expressjs/body-parser/pull/326).

I think it could be solved by increasing `arrayLimit`, because it looks like with current options sparse arrays are not allowed anyway.

@ljharb could you look into this?

PR #689, opened 2026-01-04
deps: qs@^6.14.1

UlisesGascon:
Related to https://github.com/expressjs/body-parser/issues/687

### DIFF
--- a/lib/types/urlencoded.js
+++ b/lib/types/urlencoded.js
@@ -96,7 +96,7 @@ function createQueryParser (options) {
       })
     }
 
-    var arrayLimit = extended ? Math.max(100, paramCount) : 0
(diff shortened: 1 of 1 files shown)


lib/types/urlencoded.js
```diff
@@ -96,7 +96,7 @@ function createQueryParser (options) {
       })
     }
 
-    var arrayLimit = extended ? Math.max(100, paramCount) : 0
+    var arrayLimit = extended ? Math.max(100, paramCount) : paramCount
```
jonchurch: the `Math.max` is here for extended mode to allow sparse arrays, so far as I can tell

jonchurch (approved):

Issue #258, opened 2017-08-24
can body-parser parse formData with Ajax?


sunq0001:
i tried to use bodyParser  to parse formData sent by Ajax, but req.body only returns an empty object.
does that mean bodyParser doesn't have this capability?

client.html
```html
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <meta http-equiv="X-UA-Compatible" content="ie=edge">
    <title>Document</title>
    <script>
        function submitData() {
            var form = document.getElementById('form1');
            var formData = new FormData(form);
            var xhr = new XMLHttpRequest();
            xhr.open('post', 'indexFormData.html', true);
            xhr.onload = function (e) {
                if (this.status == 200) {
                    document.getElementById('result').innerHTML = this.response;
                }
            };
            xhr.send(formData);
        }
    </script>
</head>

<body>
    <form id="form1" enctype="multipart/form-data">
        firstname:
        <input type="text" name="firstname" id="txtFirstName" value="ss"></br>
        lastname:
        <input type="text" name="lastname" id="txtLastName" value="ll"></br>
        <input type="button" value="submit" onclick="submitData()" />
    </form>
    <output id="result"></output>
</body>

</html>
```

server.js

```js
var express = require('express');
var bodyParser = require('body-parser');
var app = express();
app.use(bodyParser.urlencoded({ extended: false}));
app.use(bodyParser.text());

app.get('/indexFormData.html', function (req, res) {
    res.sendFile(__dirname + '/indexFormData.html');
});

app.post('/indexFormData.html', function (req, res) {
    res.send(req.body);
});
app.listen(1338);
```

result: {}

sunq0001:
it is too sad, i have to use multer to solve this problem. but the way it is used is quite uncomfortable.
```
var multer = require('multer')().single();
app.use(multer);
```

dougwilson:
Hi @sunq0001 that's correct: this module does not (currently) support this feature. We have the tracking issue on this as #88 . You're always welcome to implement support as well :)

PR #694, opened 2026-01-06
feat: add multipart/form-data parser support

### LINKED ISSUE #88: Add multipart/form-data support
This is just such a basic type of the web, it's hard to keep ignoring it. Because of the pattern of this module, though, we really cannot support files. But I don't see why we can support it, but just drop files. Thoughts?

Ayoub-Mabrouk:
Add multipart/form-data support

Closes #88

This PR implements multipart/form-data parsing support as requested in issue #88. The parser extracts text fields and automatically drops file fields, following the design pattern discussed by @dougwilson in the issue.

Changes

Added lib/types/multipart.js - Multipart parser implementation
Added test/multipart.js - Test suite with 12 active tests
Updated index.js - Added multipart export with lazy loading
Updated README.md - Added multipart parser documentation and examples

Implementation details

The parser uses the existing read utility, following the same pattern as other parsers in the codebase. No external dependencies are required.

It extracts text fields and drops file fields (fields with filename= in Content-Disposition header). The parser validates individual field sizes, not just overall body size, and supports per-field verification callbacks. Duplicate field names are handled by converting to arrays.

The code follows the same architecture as json, urlencoded, text, and raw parsers, with helper functions extracted for clarity: extractBoundary(), parsePart(), and addField().

Test results

All 275 tests are passing. One test is skipped due to a known Node.js stream limitation (explained below).

Skipped test explanation

The test "should handle consumed stream" is skipped because of Node.js stream semantics.

When req.resume() is called, the stream may still contain buffered data that getBody() can successfully read. There is no reliable API in Node.js to detect if a stream was previously consumed. The readableEnded, readable, and onFinished properties are not guaranteed to reflect consumption state.

Attempting to parse buffered data is correct behavior and matches the behavior of raw-body used throughout body-parser. The standard read utility doesn't handle this case either, as it relies on getBody() to handle stream consumption, which it does correctly when data is truly unavailable.

This is a known limitation of Node.js streams, not a parser bug. The parser correctly handles all real-world scenarios where streams are consumed properly.

Performance

Time complexity is O(n) where n is the body size, which is optimal since we must read the entire body. Space complexity is O(n) for body storage. The algorithm uses a single pass through the data.

Usage

```javascript
const bodyParser = require('body-parser')

// Parse multipart/form-data (text fields only, files dropped)
app.use(bodyParser.multipart())

// Or with options
app.use(bodyParser.multipart({
  limit: '1mb',        // Per-field limit
  verify: (req, res, buf, encoding) => {
    // Verify each field individually
  }
}))
```

Files changed

lib/types/multipart.js (new)
test/multipart.js (new)
index.js (added multipart export)
README.md (updated documentation)

Related issues

Closes #88 (opened Mar 29, 2015) - Original request for multipart/form-data support
Addresses #258 (opened Aug 24, 2017) - User unable to parse FormData sent via Ajax, getting empty req.body object. This PR solves that exact use case.

This implementation follows the design discussed by @dougwilson: "integrate one of those parsers into this module as lib/types/multipart.js" and the requirement to "drop the file fields and accumulate the text fields into an object". It solves the common use case where users need to extract text fields (like CSRF tokens) from multipart forms without needing external libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,9 @@ and `toString` may not be a function and instead a string or other user input.
 
 [Learn about the anatomy of an HTTP transaction in Node.js](https://nodejs.org/en/learn/http/anatomy-of-an-http-transaction).
 
-_This does not handle multipart bodies_, due to their complex and typically
-large nature. For multipart bodies, you may be interested in the following
-modules:
+_This module provides basic multipart/form-data support for text fields only._
+File fields are automatically dropped. For full file upload support, you may be
+interested in the following modules:
 
   * [busboy](https://www.npmjs.com/package/busboy#readme) and
     [connect-busboy](https://www.npmjs.com/package/connect-busboy#readme)
@@ -33,6 +33,7 @@ modules:
 This module provides the following parsers:
 
   * [JSON body parser](#bodyparserjsonoptions)
+  * [Multipart/form-data body parser](#bodyparsermultipartoptions)
   * [Raw body parser](#bodyparserrawoptions)
   * [Text body parser](#bodyparsertextoptions)
   * [URL-encoded form body parser](#bodyparserurlencodedoptions)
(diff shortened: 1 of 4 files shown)


bjohansebas (changes requested):
I haven’t reviewed it yet, but why not just continue with https://github.com/expressjs/body-parser/pull/606 instead of adding support for this format? If someone needs that format without a file, they can create it themselves, and we wouldn’t have to maintain it.

Ayoub-Mabrouk:
> I haven’t reviewed it yet, but why not just continue with #606 instead of adding support for this format? If someone needs that format without a file, they can create it themselves, and we wouldn’t have to maintain it.

Thanks for the feedback @bjohansebas . I understand PR #606 implements a generic parser interface (addressing #22) that would let users provide their own parsing functions.

I see the distinction: PR #606 provides a generic interface for custom parsers, while this PR adds multipart as a built-in parser (like json, urlencoded, text, raw already exist).

This PR addresses issue #88, which @dougwilson opened in 2015 and explicitly requested multipart/form-data support. In that issue, Doug suggested integrating a parser as `lib/types/multipart.js` - which is what this PR does. It also addresses issue #258, where users need to extract text fields (like CSRF tokens) from multipart forms.

The question is: should multipart be a built-in parser (consistent with json, urlencoded, etc.), or should users use the generic parser with their own multipart parsing function?

I'm happy to adapt based on maintainer preference. Should I close this PR in favor of the generic approach, or would both approaches coexist (generic for custom parsers, built-in multipart for convenience)?

Could you please review issue #88 (where @dougwilson explicitly requested this feature) and this PR to see if it aligns with the project's direction? I'd appreciate your thoughts on whether multipart should be built-in like the other parsers, or if the generic parser approach is the preferred solution going forward.

PR #697, opened 2026-01-16
Improve ESM compatibility

### LINKED ISSUE #666: no named exports for ESM projects
## Environment information

**Version**:
<!-- 
To find the installed version of a package, you can check the `package.json` file in the root directory of your project. The version will be listed under `dependencies` or `devDependencies`, like this:

```json
"dependencies": {
  "express": "4.18.2"
}
-->
2.2.0

**Platform**:
<!-- 
UNIX: output of uname -a
Windows: output of "$([Environment]::OSVersion.VersionString) $(('x86', 'x64')[[Environment]::Is64BitOperatingSystem])" in PowerShell console 
-->

**Node.js version**:
<!-- 
Output of node -v.
-->
20.19.2

**Any other relevant information**:

## What steps will reproduce the bug?

```mjs
import { json } from 'body-parser';
```

```
 SyntaxError: The requested module 'body-parser' does not provide an export named 'json'
```

Named exports are found by static analysis:

https://nodejs.org/en/learn/modules/publishing-a-package

> Node.js detects the named exports in CJS via [static analysis that look for certain patterns](https://github.com/nodejs/cjs-module-lexer/tree/main?tab=readme-ov-file#parsing-examples), which the example above evades. To make the named exports detectable, do this:

`defineProperty` is a supported syntax, but a `get()` method can't be a `require()` call or a call to any other function, it can only return a simple 'identifier or member expression.'

https://github.com/expressjs/body-parser/blob/d96b63da8d7445de317736471633bac83ec76cbb/index.js#L30-L34

https://github.com/nodejs/cjs-module-lexer/blob/main/README.md#getter-exports-parsing

> To avoid matching getters that have side effects, any getter for an export name that does not support the forms above will opt-out of the getter matching:

Phillip9587:
Ref: https://github.com/expressjs/body-parser/issues/666#issuecomment-3662374724

This PR contains 2 commits:

**fix: use static exports instead of lazy getters** 
Static exports allow Node.js `cjs-module-lexer` to detect named
exports correctly, improving ESM interop and bundler behavior.

**feat: add subpath exports for individual parsers**

Add an exports field to package.json exposing json, raw, text,
and urlencoded parsers as subpath exports.

This allows consumers to import individual parsers directly:
- CommonJS: require('body-parser/json')
- ESM: import json from 'body-parser/json'

The change is fully backward compatible and enables users to
opt into loading only the parsers they need.

Closes #666

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,7 @@ unreleased
 ========================
 
 * fix: use static exports instead of lazy getters to improve ESM compatibility
+* feat: add subpath exports for individual parsers
 
 2.2.2 / 2026-01-07
 =========================
--- a/README.md
+++ b/README.md
@@ -51,7 +51,14 @@ $ npm install body-parser
 ## API
 
 ```js
+// Import all parsers
 const bodyParser = require('body-parser')
+
+// Or import individual parsers directly
+const json = require('body-parser/json')
+const urlencoded = require('body-parser/urlencoded')
+const raw = require('body-parser/raw')
+const text = require('body-parser/text')
 ```
 
 The `bodyParser` object exposes various factories to create middlewares. All
--- a/package.json
+++ b/package.json
@@ -12,6 +12,18 @@
     "type": "opencollective",
     "url": "https://opencollective.com/express"
   },
+  "exports": {
+    ".": "./index.js",
+    "./package.json": "./package.json",
+    "./json": "./lib/types/json.js",
+    "./raw": "./lib/types/raw.js",
+    "./text": "./lib/types/text.js",
+    "./urlencoded": "./lib/types/urlencoded.js",
+    "./lib/*": "./lib/*.js",
+    "./lib/*.js": "./lib/*.js",
+    "./lib/types/*": "./lib/types/*.js",
+    "./lib/types/*.js": "./lib/types/*.js"
+  },
   "dependencies": {
     "bytes": "^3.1.2",
     "content-type": "^1.0.5",


jonchurch (changes requested):
Im concerned here that the exports inclusion makes this breaking to a small number of users. I am always wary of packaging changes mid major (ala files, exports), as it can break things unless the paths cover everything that was possible previously.

There are indeed apps in the wild doing silly things like importing from lib which this change would break:
https://github.com/search?q=%22require%28%27body-parser%2F%22&type=code
(I think VSCode's auto-complete for imports often does these deep imports when there's no exports field, i've seen it a lot at a previous job. likely how folks ended up w/ this)

We can split out the index.js change as it's own PR, as that's standalone and won't break anyone. Or we can do an exhaustive exports field inside this same major. ~But that doesn't seem valuable here.~

From the [Node docs](https://nodejs.org/api/packages.html#package-entry-points), we could do this to preserve back compat:

```json
  "exports": {
    ".": "./index.js",
    "./package.json": "./package.json",
    "./json": "./lib/types/json.js",
    "./raw": "./lib/types/raw.js",
    "./text": "./lib/types/text.js",
    "./urlencoded": "./lib/types/urlencoded.js",
    // Back compat entries start here
    "./lib/*": "./lib/*.js",
    "./lib/*.js": "./lib/*.js",
    "./lib/types/*": "./lib/types/*.js",
    "./lib/types/*.js": "./lib/types/*.js"
  }
```

jonchurch (approved):

Phillip9587:
@bjohansebas Should we also document the subpath exports in `README.md`? It allows to import only a specific parser like `require('body-parser/text')`.

Phillip9587:
@bjohansebas Can you review the changes please?

PR #557, opened 2024-10-31
Refactor getCharset function for conciseness and modern syntax

Ayoub-Mabrouk:
- Utilized optional chaining to simplify charset retrieval
- Improved JSDoc comments for clarity
- Enhanced readability while maintaining functionality

PR #707, opened 2026-02-23
refactor: move getCharset helper to only usage

Phillip9587:
...and uses optional chaining. In theory it short-circuits faster and by that avoids the empty string allocation and `.toLowerCase()` call. But this is only in theory and probably highly optimized in engines.

Same ideas as @Ayoub-Mabrouk in #557.

### DIFF
--- a/lib/read.js
+++ b/lib/read.js
@@ -17,7 +17,7 @@ const iconv = require('iconv-lite')
 const onFinished = require('on-finished')
 const zlib = require('node:zlib')
 const hasBody = require('type-is').hasBody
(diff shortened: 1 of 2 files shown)


bjohansebas (commented):
Is this still relevant, @Phillip9587? It doesn't seem like it is anymore.

Phillip9587:
It's not that relevant anymore as we upgraded to `content-type` v2 and updated the method. But the helper is only used in `read.js` so we should still move it there as it is not a shared helper. It was originally moved there in #551 when it was still a shared helper shared by all parsers.

PR #716, opened 2026-03-30
fix: correct off-by-one error in parameterCount

### LINKED ISSUE #715: Off-by-one in 1.x urlencoded `parameterCount` breaks parsing arrays
Discovered in https://github.com/expressjs/express/pull/7135.

[`parameterCount`](https://github.com/expressjs/body-parser/blob/b849bd533d8b4abf5576a3e301f28d9befa05ddd/lib/types/urlencoded.js#L207-L221) has an off-by-one bug (returns `0` for `"a=x"`, `1` for `"a=x&b=y"`, ...). This in turn causes an array like `"a[0]=1&a[1]=2&...&a[199]=200"` to be parsed by `qs` with `arrayLimit: 199`.
This used to work, because `arrayLimit: 199` would allow the index to be at most `199`. **`qs@6.14.2` changed `arrayLimit` behaviour** to instead restrict the array length (https://github.com/ljharb/qs/issues/540#issuecomment-3888244764, https://github.com/ljharb/qs/commit/cfc108f662326d6ab540f3545ef0b832baf83cdf, [6.14.2 changelog](https://github.com/ljharb/qs/blob/d9b4c66303375493c68c42d68e363e50b1753771/CHANGELOG.md?plain=1#L7)).

`2.x` uses a slightly [different function](https://github.com/expressjs/body-parser/blob/2322e111cc321413ec2b7b76d01be533d3de9d7d/lib/types/urlencoded.js#L129-L138) and returns the real number of parameters.

More context on the differences: https://github.com/expressjs/body-parser/pull/652#issuecomment-3575640861

## Environment information

**Version**: 1.20.4

**Platform**: Microsoft Windows NT 10.0.19045.0 x64

**Node.js version**: 25.8.2

**Any other relevant information**: When using public API, the problem can be observed only with `qs >=6.14.2`

## What steps will reproduce the bug?

1. Try parsing a large (>100 elements) array using extended urlencoded parser

Alternatively:

1. Run `npm test`, the following test will fail:
   https://github.com/expressjs/body-parser/blob/b849bd533d8b4abf5576a3e301f28d9befa05ddd/test/urlencoded.js#L167-L185

abhu85:
## Summary

Fix off-by-one error in `parameterCount` that causes large arrays to be parsed as objects with qs@6.14.2.

## Problem

The `parameterCount` function counts `&` characters but returns that count directly. Since the number of parameters equals the number of `&` characters plus one, this causes an off-by-one error.

For example, with `a[0]=1&a[1]=2&...&a[199]=200` (200 parameters):
- Before fix: returned 199 (count of `&` chars)
- After fix: returns 200 (actual parameter count)

This became problematic with [qs@6.14.2](https://github.com/ljharb/qs/commit/cfc108f662326d6ab540f3545ef0b832baf83cdf) which changed `arrayLimit` behavior to restrict array length instead of max index.

## Solution

Change `return count` to `return count + 1` in `parameterCount`.

## Test Plan

- [x] Existing "should parse array index notation with large array" test verifies fix
- [x] All 262 tests pass
- [x] Linting passes

## Compatibility

- No breaking changes
- Backwards compatible (arrays that worked before still work)

Fixes #715

### DIFF
--- a/test/urlencoded.js
+++ b/test/urlencoded.js
@@ -506,6 +506,31 @@ describe('bodyParser.urlencoded()', function () {
           .expect(expectKeyCount(10000))
           .expect(200, done)
       })
+
+      it('should correctly count parameters for array parsing', function (done) {
+        // Test for off-by-one bug fix (issue #715)
+        // With 110 array elements, there are 110 parameters (109 & chars)
+        // Before fix: parameterCount returned 109, arrayLimit was 109
+        // After fix: parameterCount returns 110, arrayLimit is 110
+        var server = createServer({ extended: true, parameterLimit: 200 })
+        var arrayParams = Array.from({ length: 110 }, function (_, i) {
+          return 'a[' + i + ']=' + (i + 1)
+        }).join('&')
+
+        request(server)
+          .post('/')
+          .type('form')
+          .send(arrayParams)
+          .expect(function (res) {
+            // The body should contain an array 'a' with 110 elements
+            // If parameterCount returns 109 (bug), arrayLimit would be 109,
+            // and qs would convert indices >= 109 to object keys
+            var body = JSON.parse(res.text)
+            assert.ok(Array.isArray(body.a), 'should have array "a", got: ' + res.text.substring(0, 200))
+            assert.strictEqual(body.a.length, 110, 'array should have 110 elements')
+          })
+          .expect(200, done)
+      })
     })
   })
 
--- a/test/urlencoded.js
+++ b/test/urlencoded.js
@@ -506,31 +506,6 @@ describe('bodyParser.urlencoded()', function () {
           .expect(expectKeyCount(10000))
           .expect(200, done)
       })
-
-      it('should correctly count parameters for array parsing', function (done) {
-        // Test for off-by-one bug fix (issue #715)
-        // With 110 array elements, there are 110 parameters (109 & chars)
-        // Before fix: parameterCount returned 109, arrayLimit was 109
-        // After fix: parameterCount returns 110, arrayLimit is 110
-        var server = createServer({ extended: true, parameterLimit: 200 })
-        var arrayParams = Array.from({ length: 110 }, function (_, i) {
-          return 'a[' + i + ']=' + (i + 1)
-        }).join('&')
-
-        request(server)
-          .post('/')
-          .type('form')
-          .send(arrayParams)
-          .expect(function (res) {
-            // The body should contain an array 'a' with 110 elements
-            // If parameterCount returns 109 (bug), arrayLimit would be 109,
-            // and qs would convert indices >= 109 to object keys
-            var body = JSON.parse(res.text)
-            assert.ok(Array.isArray(body.a), 'should have array "a", got: ' + res.text.substring(0, 200))
-            assert.strictEqual(body.a.length, 110, 'array should have 110 elements')
-          })
-          .expect(200, done)
-      })
     })
   })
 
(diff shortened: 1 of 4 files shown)


test/urlencoded.js
```diff
@@ -506,6 +506,31 @@ describe('bodyParser.urlencoded()', function () {
...
+          .expect(function (res) {
+            // The body should contain an array 'a' with 110 elements
+            // If parameterCount returns 109 (bug), arrayLimit would be 109,
+            // and qs would convert indices >= 109 to object keys
+            var body = JSON.parse(res.text)
+            assert.ok(Array.isArray(body.a), 'should have array "a", got: ' + res.text.substring(0, 200))
+            assert.strictEqual(body.a.length, 110, 'array should have 110 elements')
+          })
+          .expect(200, done)
+      })
```
krzysdz: There already is a test that verifies this

https://github.com/expressjs/body-parser/blob/b849bd533d8b4abf5576a3e301f28d9befa05ddd/test/urlencoded.js#L167-L185

abhu85:
Thanks for pointing that out. I see the existing test at L167-185 with 500 elements.

I added my test because I wanted to verify the specific off-by-one fix, but you're right that the existing test should theoretically catch this too since:
- 500 parameters → paramCount returns 499 (bug)
- arrayLimit = Max.max(100, 499) = 499
- qs 6.14.2+ would convert index 499 to object key

Does the existing test currently fail on main without this fix? If so, I'm happy to remove my test as redundant. If there's something causing it to pass despite the bug, my test provides additional coverage.

abhu85:
Friendly ping on this PR. I asked about the test coverage question on March 31st - just wanted to check if you had a chance to verify whether the existing L167-185 test catches this bug without my fix, or if my additional test provides unique coverage.

Happy to remove the test if it's redundant, or keep it if it adds value. Thanks!

krzysdz:
>  just wanted to check if you had a chance to verify whether the existing L167-185 test catches this bug without my fix

You can check it yourself. Just run tests without your changes and check if the test in question fails. When submitting bugfixes, please also make sure that the new regression tests fail without the patch (in doing so you'd notice existing tests failing too).

By the way, in #715 I pointed out that this was found because of a failing test in Express and the exact same test is a part of `body-parser` test suite.

abhu85:
You're absolutely right, thank you for the guidance.

I verified on the 1.x branch:

```
# Without fix
npm test
202 passing
1 failing

  1) bodyParser.urlencoded()
       with extended option
         when true
           should parse array index notation with large array:
      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal

# With fix (return count + 1)
npm test
262 passing
```

The existing "should parse array index notation with large array" test at L235-252 does catch this bug - it sends 500 array elements (f[0] through f[499]) and expects `obj.f.length === 500`.

Since the existing test provides adequate regression coverage, I'll remove my redundant test and push an updated commit. Thanks for pointing this out!

abhu85:
Done - pushed [f49cd5f](https://github.com/abhu85/body-parser/commit/f49cd5f) removing the redundant test. The PR now contains only the one-line fix.

All 262 tests pass.

jonchurch (approved):
Hey thank you @abhu85  and @krzysdz !

I pushed commits here to:

- Update to qs@~6.14.2, with this change we now need to raise the qs version floor we support.
- Refactor this to use the logic we have on the 2.x line, which returns the _actual count_ of parameters vs parameterCount - 1 (which was correct in 1.x while qs was looking for parameter count - 1)
- Updated HISTORY.md

krzysdz (approved):

PR #728, opened 2026-05-11
Upgrade "content-type"

blakeembrey:
The parser in `content-type@2` is 3x faster and doesn't throw anymore.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -27,7 +27,7 @@ module.exports = {
 function getCharset (req) {
   var header = req.headers['content-type']
   if (!header) return undefined
-  return (contentType.parse(header).parameters.charset || '').toLowerCase()
+  return contentType.parse(header).parameters.charset?.toLowerCase()
 }
 
 /**
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -25,11 +25,9 @@ module.exports = {
  * @private
  */
 function getCharset (req) {
-  try {
-    return (contentType.parse(req).parameters.charset || '').toLowerCase()
-  } catch {
-    return undefined
-  }
+  var header = req.headers['content-type']
+  if (!header) return ''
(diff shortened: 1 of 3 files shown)


Phillip9587 (approved):
LGTM! But we need to also update it in `type-is` and `express` to not duplicate the module in the dependency tree: https://npmgraph.js.org/?q=express#select=content-type%401.0.5

blakeembrey:
😓 Oh damn I missed it was used in this package, since it's still support node 0.6 and has a hard API to work with I guess it's a good candidate to start refactoring. If/when I release the new `type-is` should I bump it in this PR or a new one?

Phillip9587:
I think it does not matter and it is up to you if you but it in a separate PR. I would prefer a single PR for body-parser and one for express.

blakeembrey:
@Phillip9587 Released `type-is@2.1.0` with `content-type@2`, bumped the version of `type-is` in this PR.

lib/utils.js
```diff
@@ -25,11 +25,9 @@ module.exports = {
  * @private
  */
 function getCharset (req) {
-  try {
-    return (contentType.parse(req).parameters.charset || '').toLowerCase()
-  } catch {
-    return undefined
-  }
+  var header = req.headers['content-type']
+  if (!header) return ''
```
Phillip9587:
```suggestion
  if (!header) return undefined
```

lib/utils.js
```diff
@@ -25,11 +25,9 @@ module.exports = {
...
+  return (contentType.parse(header).parameters.charset || '').toLowerCase()
```
Phillip9587:
```suggestion
  return contentType.parse(header).parameters.charset?.toLowerCase()
```

Phillip9587 (changes requested):

Phillip9587 (approved):
LGTM. Thanks for all the work @blakeembrey

PR #742, opened 2026-07-06
fix: avoid mutating read options when reading body

Phillip9587:
The parser read options were created once per middleware instance and reused for each request. `read()` was aliasing that shared object and adding per-request `length` and `encoding` fields before passing it to `raw-body`.

Pass `raw-body` an explicit `rawBodyOptions` object containing the per-request values it needs.

### DIFF
--- a/lib/read.js
+++ b/lib/read.js
@@ -79,38 +79,37 @@ function read (req, res, next, parse, debug, options) {
   }
 
   let length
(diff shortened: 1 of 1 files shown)


UlisesGascon (approved):

bjohansebas (approved):
LGTM, thanks!

PR #752, opened 2026-07-23
docs: note that req.body may be undefined in v2

### LINKED ISSUE #632: Migration Guide V1 -> V2
### Description

For an empty http request body the v1 json parser would provide a {} request body. V2 provides an `undefined` body

### Expectations

This to be described as it can cause all kinds of issues and errors. If you have done that and I missed it, sorry.

If you expect for example a username entry:

```js
if (req.body.username){
   // do something
}
```

It has to change to something like this

```js
if (req.body?.username){
   // do something
}
```

I added a workaround to not have to change lot's of code

```js
const bodyParser        = require('body-parser'),
      express           = require('express'),
      json_reviver      = require('../../lib/json_reviver')

function jsonWithDefault(options) {
  const jsonParser = bodyParser.json(options);
  return (req, res, next) => {
    jsonParser(req, res, (err) => {
      if (err) return next(err);
      if (req.body === undefined) {
        req.body = {};   // restore v1 behavior
      }
      next();
    });
  };
}

module.exports = {
  jsonParser: jsonWithDefault({ limit: '5mb', reviver: json_reviver }),
  urlEncoded: bodyParser.urlencoded({ extended: true, limit: '5mb' }),
  raw:   express.raw({
    inflate: true,
    limit: '50mb',
    type: () => true, // this matches all content types
  })
};

```

Hashim1999164:
## Summary
- Document the v1 to v2 behavior change where `req.body` stays `undefined` when no body is parsed, instead of always being initialized to `{}`.
- Call out optional chaining / defaults for code that assumed an object.

Fixes #632

## Test plan
- [ ] Confirm the new note reads clearly in the README API section
- [ ] Confirm it matches the breaking change already listed in HISTORY.md for 2.0.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,12 @@ The `bodyParser` object exposes various factories to create middlewares. All
 middlewares will populate the `req.body` property with the parsed body when
 the `Content-Type` request header matches the `type` option.
 
+**Note** As of v2, `req.body` is left `undefined` when no body is parsed (for
+example an empty body or a non-matching `Content-Type`). In v1, `req.body` was
+always initialized to `{}`. Code that assumes `req.body` is always an object
+should use optional chaining (for example `req.body?.username`) or set a
+default.
+
 The various errors returned by this module are described in the
 [errors section](#errors).
 


kilisamemarisaaa (commented):
The "empty body" example is broader than v2's actual behavior. A request with a matching `Content-Type: application/json` and `Content-Length: 0` is still considered to have a body, and the JSON parser returns `{}`; the existing `test/json.js` cases "should handle Content-Length: 0" and "should handle empty message-body" both assert that result.

The breaking change applies when no body is parsed, for example when the request has neither `Content-Length` nor `Transfer-Encoding`, or when `Content-Type` does not match. Could this example be narrowed so it does not imply that all zero-length JSON requests now leave `req.body` undefined?

Hashim1999164:
yeah my bad, empty json with content length 0 still parses to {}. i narrowed the note to the no body parsed cases, missing content length or transfer encoding, or a type that doesnt match.

PR #754, opened 2026-08-02
Fix charset decoding for text bodies

### LINKED ISSUE #680: Almost all non-utf8 encodings are incorrect
Even the most basic `windows-1252` which `latin1` and `ascii` alias to:
```js
const express = require('express')
const bodyParser = require('body-parser')

const app = express()
app.use(bodyParser.urlencoded())
app.use(bodyParser.json())
app.use(bodyParser.text())

app.use(function (req, res) {
  res.setHeader('Content-Type', 'text/plain')
  res.write('you posted:\n')
  res.write(`${escape(req.body)}\n`)
  res.end(String(req.body))
})

app.listen(8080, async () => {
  const res = await fetch('http://localhost:8080/', {
    method: 'POST',
    headers: { 'content-type': 'text/plain; charset=windows-1252' },
    body: Uint8Array.of(0x80, 0x81, 0x82, 0x83, 0x8d, 0x9e, 0x9f)
  })

  console.log(await res.text())
})
```

Results in:
```
you posted:
%u20AC%uFFFD%u201A%u0192%uFFFD%u017E%u0178
€�‚ƒ�žŸ
```

But it should be `€\x81‚ƒ\x8DžŸ` instead, with no replacement chars
i.e. `%u20AC%81%u201A%u0192%8D%u017E%u0178` escaped

See Encoding Standard: https://encoding.spec.whatwg.org/
All characters are mapped in https://encoding.spec.whatwg.org/index-windows-1252.txt, including `0x81` and `0x8D`.

Same goes for other encodings: half of single-bytes are mapped incorrectly and contradict the spec: all of `windows-*` family except `windows-1256`, `koi8-u`, `macintosh`.

---

All of legacy multi-bytes that are supported also behave incorrectly

---

UTF-16 also behaves incorrectly:

```js
const express = require('express')
const bodyParser = require('body-parser')

const app = express()
app.use(bodyParser.urlencoded())
app.use(bodyParser.json())
app.use(bodyParser.text())

app.use(function (req, res) {
  res.setHeader('Content-Type', 'text/plain')
  res.write('you posted:\n')
  res.write(`Is well formed: ${req.body.isWellFormed()}\n`)
  res.write(`${escape(req.body)}\n`)
  res.end(String(req.body))
})

app.listen(8080, async () => {
  const res = await fetch('http://localhost:8080/', {
    method: 'POST',
    headers: { 'content-type': 'text/plain; charset=utf-16le' },
    body: Uint8Array.of(0, 0xd8, 0, 0xd8)
  })

  console.log(await res.text())
})
```

Results in:
```
you posted:
Is well formed: false
%uD800%uD800
��
```

But per spec it should never produce non-well-formed strings and should instead have produced replacements chars, i.e. `%uFFFD%uFFFD` escaped

See spec: https://encoding.spec.whatwg.org/#shared-utf-16-decoder

This could have potential security impact

These decoders are enabled in the default configuration
The default utf-8 decoder never produces non-well-formed strings, but the client can force that by specifying utf-16 encoding, while per spec that shouldn't be possible (produced strings should be always well-formed)

deepakganesh78:
Fixes #680

## Reproduction
A `text/plain; charset=windows-1252` request containing bytes `80 81 82 83 8d 9e 9f` decoded to `€�‚ƒ�žŸ`, replacing C1 control mappings at `0x81` and `0x8d`. A `text/plain; charset=utf-16le` request containing two unpaired high surrogates decoded to a non-well-formed JavaScript string instead of replacement characters.

## Root cause
`raw-body`/`iconv-lite` decoding is used for request character sets. For encodings covered by the WHATWG Encoding Standard, that can differ from the platform `TextDecoder` mappings and UTF-16 replacement behavior available in supported Node.js versions.

## Fix
Read parser bodies as bytes, then decode with Node's WHATWG `TextDecoder` when it supports the charset. Keep `iconv-lite` as the fallback for legacy charsets that `TextDecoder` does not support, and keep generic `utf-16` on the existing iconv-lite path to preserve current BOM-based behavior.

## Compatibility notes
This intentionally changes decoded text for WHATWG-supported charset labels to match `TextDecoder`; unsupported labels and existing `utf-16`/`utf-32` behavior continue through `iconv-lite`.

## Validation
- Regression tests fail without the fix: `npx mocha --reporter spec --check-leaks test\text.js --grep "windows-1252|utf-16"` -> 0 passing, 2 failing.
- `npm test` -> 271 passing.
- `npm run lint` -> passed.

### DIFF
--- a/lib/read.js
+++ b/lib/read.js
@@ -15,6 +15,7 @@ const createError = require('http-errors')
 const getBody = require('raw-body')
 const iconv = require('iconv-lite')
 const onFinished = require('on-finished')
+const { TextDecoder } = require('node:util')
 const zlib = require('node:zlib')
 const hasBody = require('type-is').hasBody
 const contentType = require('content-type')
@@ -93,7 +94,7 @@ function read (req, res, next, parse, debug, options) {
   }
 
   // assert charset is supported
-  if (verify && encoding !== null && !iconv.encodingExists(encoding)) {
+  if (encoding !== null && !isEncodingSupported(encoding)) {
     return next(createError(415, 'unsupported charset "' + encoding.toUpperCase() + '"', {
       charset: encoding.toLowerCase(),
       type: 'charset.unsupported'
@@ -103,7 +104,7 @@ function read (req, res, next, parse, debug, options) {
   // set raw-body options
   const rawBodyOptions = {
     length,
-    encoding: verify ? null : encoding,
+    encoding: null,
     limit: options.limit
   }
 
@@ -155,9 +156,7 @@ function read (req, res, next, parse, debug, options) {
     let str = body
     try {
       debug('parse body')
-      str = typeof body !== 'string' && encoding !== null
-        ? iconv.decode(body, encoding)
-        : body
+      str = decodeBody(body, encoding)
       req.body = parse(str, encoding)
     } catch (err) {
       next(createError(400, err, {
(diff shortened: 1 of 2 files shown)


bjohansebas (changes requested):
I actually recommend upgrading to the latest version of raw-body instead of using iconv-lite. It also lets you override the TextDecoder, as raw-body now does, in case anyone needs support for UTF-7 or other custom encodings.

